PR #13, opened 2016-10-14
chore (doc): add tutorials for stash and github

fsamin:
Signed-off-by: François Samin <francois.samin@corp.ovh.com>


proullon:
"chore"

PR #16, opened 2016-10-17
fix: error management + return last_modified on application update

sguiheux:


engine/api/application.go
```diff
@@ -524,20 +524,22 @@ func updateApplicationHandler(w http.ResponseWriter, r *http.Request, db *sql.DB
...
 		return
 	}
 
 	var appPost sdk.Application
 	// Get body
 	data, err := ioutil.ReadAll(r.Body)
 	if err != nil {
-		w.WriteHeader(http.StatusBadRequest)
+		log.Warning("updateApplicationHandler> Cannot read body: %s\n", err)
+		WriteError(w, r, err)
```
fsamin: i suggest to keep WriteHeader, because il the body can't be read, status code should be 400 and not 500

engine/api/application.go
```diff
@@ -524,20 +524,22 @@ func updateApplicationHandler(w http.ResponseWriter, r *http.Request, db *sql.DB
...
 		return
 	}
 	err = json.Unmarshal(data, &appPost)
 	if err != nil {
-		w.WriteHeader(http.StatusBadRequest)
+		log.Warning("updateApplicationHandler> Cannot unmarshal request: %s\n", err)
+		WriteError(w, r, err)
```
fsamin: i suggest to keep WriteHeader, because il the body can't be parsed, status code should be 400 and not 500

sguiheux: Need WriteError with a message for UI.
I will create a dedicated error in sdk for this.

fsamin: Indeed, that's the best way !

engine/api/application.go
```diff
@@ -272,20 +272,22 @@ func addApplicationHandler(w http.ResponseWriter, r *http.Request, db *sql.DB, c
...
 	if err != nil {
-		w.WriteHeader(http.StatusBadRequest)
+		log.Warning("addApplicationHandler: Cannot read body: %s\n", err)
+		WriteError(w, r, sdk.ErrWrongRequest)
 		return
 	}
 	err = json.Unmarshal(data, &app)
 	if err != nil {
-		w.WriteHeader(http.StatusBadRequest)
+		log.Warning("addApplicationHandler: Cannot unmarshal requiest: %s\n", err)
```
fsamin: requiest

sguiheux:
fixed

PR #21, opened 2016-10-20
fix: return last_modified on repo attach and detach

sguiheux:


engine/api/application/application.go
```diff
@@ -52,14 +52,14 @@ func CreateFromWizard(db database.QueryExecuter, app *sdk.Application, p *sdk.Pr
 			return err
 		}
 
-		repo, err := client.RepoByFullname(app.RepositoryFullname)
+		_, err = client.RepoByFullname(app.RepositoryFullname)
```
fsamin: one line plz

sguiheux: done

PR #63, opened 2016-11-10
refactor(hatchery)

yesnault:
* use subCommand : hatchery local --api... hatchery docker --api...
* move exchanged struct to sdk
* log level arg


proullon:
nice

PR #78, opened 2016-11-28
feat: refactor keys & default value to enabled attr in step : true

yesnault:


yesnault:
hum, seems no more conflict

PR #79, opened 2016-11-28
With trigger

sguiheux:


engine/api/trigger/trigger.go
```diff
@@ -331,6 +331,60 @@ func LoadAutomaticTriggersAsSource(db database.Querier, appID, pipelineID, envID
 	return triggers, nil
 }
 
+// LoadTriggersByAppAndPipeline Load triggers for the given app and pipeline
+func LoadTriggersByAppAndPipeline(db database.Querier, appID int64, pipID int64) ([]sdk.PipelineTrigger, error) {
```
fsamin: We should have tests for functions using funky queries and scanner functions

engine/api/application.go
```diff
@@ -218,6 +219,18 @@ func getApplicationHandler(w http.ResponseWriter, r *http.Request, db *sql.DB, c
 		}
 	}
 
+	if withTriggers == "true" {
+		for i := range app.Pipelines {
+			appPip := &app.Pipelines[i];
```
fsamin: why appPip is not defined by the range ?

sguiheux: Need a pointer to set trigger on it just after

engine/api/application_test.go
```diff
@@ -0,0 +1,170 @@
...
+
+	"github.com/gorilla/mux"
+	"github.com/stretchr/testify/assert"
+
+	"github.com/ovh/cds/engine/api/application"
+	"github.com/ovh/cds/engine/api/auth"
+	"github.com/ovh/cds/engine/api/pipeline"
+	"github.com/ovh/cds/engine/api/sessionstore"
+	"github.com/ovh/cds/engine/api/testwithdb"
+	test "github.com/ovh/cds/engine/api/testwithdb"
```
fsamin: double import

PR #90, opened 2016-12-12
feat (api): mv scheduler package to queue package

fsamin:
i need scheduler package to **schedule** pipelines as cron job


fsamin:
plz review

PR #123, opened 2016-12-28
fix (cli, api): add & update template logs

fsamin:


sdk/template.go
```diff
@@ -194,6 +199,8 @@ func UploadTemplate(filePath string, update bool, name string) ([]byte, error) {
 			return nil, errjson
 		}
 
+		log.Println("Getting templates list : OK")
```
yesnault: verbose

sdk/template.go
```diff
@@ -205,6 +212,9 @@ func UploadTemplate(filePath string, update bool, name string) ([]byte, error) {
 		if !found {
 			return nil, fmt.Errorf("Template %s not found", name)
 		}
+
+		log.Printf("Found template at %s\n", path)
```
yesnault: verbose

PR #132, opened 2017-01-03
fix (api): speed up load model status func

fsamin:


engine/api/worker/elasticity.go
```diff
@@ -7,7 +7,10 @@ import (
 
 	"encoding/json"
 
+	"sync"
+
```
sguiheux: wrong place

engine/api/worker/elasticity.go
```diff
@@ -77,60 +80,119 @@ func LoadWorkerModelStatusForGroup(db *sql.DB, groupID int64) ([]sdk.ModelStatus
...
-			return nil, err
+			return err
 		}
-		status = append(status, ms)
+		defer rows.Close()
+
+		for rows.Next() {
+			var model int64
+			var count int64
+			err := rows.Scan(&model, &count)
```
sguiheux: one line

PR #149, opened 2017-01-12
feat, fix (cli, api): truncate warnings, maintenance mode, worker model handlers

fsamin:


cli/cds/admin/maintenance/cmd.go
```diff
@@ -0,0 +1,87 @@
+package maintenance
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+
+	"encoding/json"
```
sguiheux: no

PR #184, opened 2017-01-24
feat (api,cli): export and import users, groups and projects for admin only

fsamin:


cli/cds/admin/export/cmd.go
```diff
@@ -0,0 +1,146 @@
+package export
+
+import "github.com/spf13/cobra"
```
sguiheux: 1 import to rules them all

cli/cds/admin/importer/cmd.go
```diff
@@ -0,0 +1,160 @@
+package importer
+
+import (
+	"encoding/json"
+	"fmt"
+
```
sguiheux: review import

PR #190, opened 2017-01-24
feat (cli): admin users command

fsamin:


cli/cds/user/user.go
```diff
@@ -1,8 +1,15 @@
 package user
 
-import "github.com/spf13/cobra"
+import (
+	"fmt"
+
+	"github.com/ovh/cds/sdk"
```
sguiheux: wrong place

PR #197, opened 2017-01-27
feat(hatcherySwarm): force pull if tag is latest

yesnault:


engine/hatchery/swarm/swarm.go
```diff
@@ -355,7 +355,7 @@ checkImage:
 		}
 	}
 
-	if !imageFound {
+	if !imageFound || strings.HasSuffix(model.Image, ":latest") {
```
fsamin: humm, maybe we could also avoid do the ListImages()  above

yesnault: yep, fixed

yesnault: in next commit...

yesnault: fixed

PR #260, opened 2017-02-15
feat (worker): improvments

fsamin:
* Avoid check queue while building
 * Check Worker version vs. API version on registration


engine/worker/main.go
```diff
@@ -174,7 +180,10 @@ func queuePolling() {
 			os.Exit(0)
 		}
 
-		checkQueue()
+		if !building {
```
sguiheux: no need, take action is in checkQueue

PR #283, opened 2017-02-21
feat (ui): clone notification

sguiheux:


ui/src/assets/i18n/en.json
```diff
@@ -161,6 +163,7 @@
   "notification_send_initiator" : "Send to initiator",
   "notification_title" : "Notification title",
   "notification_body" : "Notification content",
+  "notification_type_title" : "Notification type: ",
```
fsamin: remove Notification on every labels

PR #318, opened 2017-03-02
feat(plugins): refactor templating

yesnault:


contrib/plugins/plugin-marathon/marathon.go
```diff
@@ -279,52 +276,36 @@ func (m MarathonPlugin) Run(a plugin.IJob) plugin.Result {
...
-		}
-	}
-
-	funcMap := template.FuncMap{
-		"title":  strings.Title,
-		"lower":  strings.ToLower,
-		"upper":  strings.ToUpper,
-		"escape": Escape,
+	// apply cds.var on marathon.json file
+	out, errapp := plugin.ApplyArgumentsOnString(a.Arguments().Data, string(buff))
```
fsamin: plugin.ApplyArguments should take a byte array as argument

PR #320, opened 2017-03-03
fix update scheduler + get date

sguiheux:


engine/api/scheduler.go
```diff
@@ -291,6 +291,15 @@ func updateSchedulerApplicationPipelineHandler(w http.ResponseWriter, r *http.Re
 		return sdk.WrapError(err, "updateSchedulerApplicationPipelineHandler> Cannot update scheduler")
 	}
 
+	nx, errN := scheduler.LoadNextExecution(tx, sOld.ID, sOld.Timezone)
```
fsamin: should lock it

PR #405, opened 2017-03-28
fix (hatcheryOpenstack): server cache deadlock

fsamin:


engine/hatchery/openstack/openstack.go
```diff
@@ -305,6 +305,8 @@ func (h *HatcheryCloud) KillWorker(worker sdk.Worker) error {
 // SpawnWorker creates a new cloud instances
 // requirements are not supported
 func (h *HatcheryCloud) SpawnWorker(model *sdk.Model, job *sdk.PipelineBuildJob) error {
+	log.Notice("SpawnWorker> %s", model.Name)
```
yesnault: with book job?

PR #493, opened 2017-04-11
fix (ui): fix color of waiting state in ui

bnjjj:
I think orange isn't good choice for "waiting" state. When I've tried the new ui with a pipeline and saw this color I thought that my pipeline was broken. Maybe blue is more appropriate ?


ui/src/app/shared/status/status.icon.html
```diff
@@ -6,5 +6,5 @@
     <i class="remove red icon" *ngSwitchCase="'Fail'"></i>
     <i class="ban grey icon" *ngSwitchCase="'Disabled'"></i>
     <i class="ban grey icon" *ngSwitchCase="'Skipped'"></i>
-    <i class="wait orange icon" *ngSwitchCase="'Waiting'"></i>
+    <i class="wait blue icon" *ngSwitchCase="'Waiting'"></i>
```
sguiheux: Must override .blue.icon ( semantic icon color) to use $cds_color_teal  => style.scss
Missing TU impact: status.spec.ts and workflow.spec.ts

yesnault: Building is blue. Waiting have to be another color. We can update for another color, but update have to be done on UI2 & CLI (cds ui, then "m")

sguiheux: I think your pb is with CLI. But in WebUI i d'ont think it's a bad idea. 
Waiting: blue clock icon
Building: blue loader

yesnault: ok. We will update cli after if necessary

PR #502, opened 2017-04-12
fix (plugin): vet & name

yesnault:


yesnault:
name -> 
"download" -> import KO
"plugin-download" -> import OK

PR #515, opened 2017-04-14
feat (template): template with a simple GitClone Step

yesnault:

### DIFF
--- a/contrib/templates/cds-template-only-git-clone-job/template.go
+++ b/contrib/templates/cds-template-only-git-clone-job/template.go
@@ -1,0 +1,115 @@
+package main
+
+import (
+	"github.com/ovh/cds/sdk"
+	"github.com/ovh/cds/sdk/template"
+)
+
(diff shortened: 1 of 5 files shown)


contrib/templates/cds-template-only-git-clone-job/template.go
```diff
@@ -0,0 +1,115 @@
...
+
+func (t *TemplateOnlyGitCloneJob) Name() string {
+	return "cds-template-only-git-clone-job"
+}
+
+func (t *TemplateOnlyGitCloneJob) Description() string {
+	return `
+This template creates:
+- a build pipeline with	one stage, containing one job
+- job contains 2 steps: CDS_GitClone and a empty script.
```
fsamin: Why not use the builtin gitclone action ? It's better than the old CDS_GitClone one

yesnault:
Yes, I'll update that

Issue #517, opened 2017-04-15
fix (ui) : delete app -> app list not refresh


yesnault:

sguiheux:
Like other refreshing issue:  cf #488
Don't need to open an issue on each refreshing ressource issue.
There will be a big refacto on UI when 488 will be fixed.
Closing this one, and updating #481

PR #524, opened 2017-04-17
fix(ui): delete app -> app list in a project is now refreshed. cf #517

bnjjj:
It works but maybe it's not the right way for you. Let me know !

#517


ui/src/app/views/application/show/admin/application.admin.component.ts
```diff
@@ -51,6 +52,8 @@ export class ApplicationAdminComponent implements OnInit {
 
     deleteApplication(): void {
         this.loading = true;
+
+        this._projectStore.deleteApplication(this.project.key, this.application.name);
```
bnjjj: Maybe it could be better if I put this line in the applicationStore ?

PR #588, opened 2017-05-04
feat (api): workflow structures, and API routes

fsamin:
cc @bnjjj 

TODO:

* WorkflowNodeJoin
* ProjectLastUpdates including workflows
* UserPermissions


engine/api/workflow/dao.go
```diff
@@ -1,22 +1,176 @@
 package workflow
 
 import (
+	"database/sql"
+	"time"
+
 	"github.com/go-gorp/gorp"
+
+	"fmt"
```
sguiheux: wrong place

PR #637, opened 2017-05-19
fix (api) : use delete cascade on workflow model

sguiheux:


engine/api/workflow.go
```diff
@@ -84,35 +83,22 @@ func postWorkflowHandler(w http.ResponseWriter, r *http.Request, db *gorp.DbMap,
...
 	if errl != nil {
-		return sdk.WrapError(errT, "Cannot load workflow")
-	}
-
-	if !detailed {
-		return WriteJSON(w, r, wf1, http.StatusCreated)
+		return sdk.WrapError(errl, "Cannot load workflow")
 	}
 
-	w2 := sdk.DetailedWorkflow{}
```
fsamin: remove DetailedWorkflow in the sdk package

engine/api/workflow/dao_test.go
```diff
@@ -361,12 +361,13 @@ func TestUpdateSimpleWorkflowWithApplicationEnvPipelineParametersAndPayload(t *t
...
 	t.Logf("Modifying workflow... with %d instead of %d", app2.ID, app.ID)
 
+	w1old := w1
 	w1.Name = "test_2"
 	w1.Root.PipelineID = pip2.ID
 	w1.Root.Context.Application = &app2
 	w1.Root.Context.ApplicationID = app2.ID
 
-	test.NoError(t, Update(db, w1, u))
+	test.NoError(t, Update(db, w1, w1old, u))
```
fsamin: i think w1 is a pointer to a workflow, so this is not revelant

PR #638, opened 2017-05-20
fix (ui): error hide when no group added

Hugome:
On project creation if no group added, the error not displayed.

I add a error `div` below "Add an existing group" which is displayed when no group added, and hide when there is groups.

### DIFF
--- a/ui/src/app/views/project/add/project.add.html
+++ b/ui/src/app/views/project/add/project.add.html
@@ -36,15 +36,19 @@
                     <h3>{{ 'project_permission_list_title' | translate }}</h3>
                     <app-permission-list [permissions]="project.groups" mode="form"
                                          (event)="permissionManagement($event)"></app-permission-list>
-                    <div class="ui error message" *ngIf="groupError">
-                        {{ 'project_groups_error' | translate }}
-                    </div>
                 </div>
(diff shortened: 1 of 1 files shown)


ui/src/app/views/project/add/project.add.html
```diff
@@ -47,6 +47,13 @@
                     <app-permission-form (createGroupPermissionEvent)="permissionManagement($event)" buttonType="button" #permForm></app-permission-form>
                 </div>
             </div>
+            <div class="row" *ngIf="(!project.groups || project.groups.length <= 0) && groupError">
+                <div class="column">
+                    <div class="ui error message">
+                        {{ 'project_groups_error' | translate }}
+                    </div>
+                </div>
+            </div>
```
yesnault: no need to remove div line 39?

Hugome: In fact the error position change if there are group or not  
No group
image
Groups but with other errors
image
But i can remove the first div and keep only the new

sguiheux: +1 to remove the first one

Hugome: It is done :)

Issue #653, opened 2017-05-23
fix (ui): improve search application/project in navbar


sguiheux:
Need to implement something like fuzzy filter

sguiheux:
can use fullTextSearch on dropdown
https://semantic-ui.com/modules/dropdown.html#/settings

PR #663, opened 2017-05-24
fix (ui): fix issue #653 and an refactor workflow.component

bnjjj:

### DIFF
--- a/ui/src/app/views/application/show/workflow/application.workflow.component.ts
+++ b/ui/src/app/views/application/show/workflow/application.workflow.component.ts
@@ -30,7 +30,7 @@ export class ApplicationWorkflowComponent implements OnInit {
 
     // Filter values
     branches: Array<Branch>;
-    versions: Array<number>;
+    versions: Array<string>;
 
     // Modal Component to link pipeline
     @ViewChild('linkPipelineComponent')
@@ -199,7 +199,7 @@ export class ApplicationWorkflowComponent implements OnInit {
     /**
      * Action when changing version
      */
-    changeVersion(version?: number): void {
+    changeVersion(version?: string): void {
         this.applicationFilter.branch = this.applicationFilter.branch.trim();
 
         if (!version && Array.isArray(this.versions) && this.versions.length) {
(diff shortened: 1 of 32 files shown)


ui/src/app/views/application/show/workflow/application.workflow.component.ts
```diff
@@ -193,29 +191,39 @@ export class ApplicationWorkflowComponent implements OnInit {
...
     }
 
+    /**
+     * Load the list of version for the current application on the selected branch
+     */
+    loadVersions(key: string, appName: string): Observable<Array<number>> {
+        return this._appWorkflow.getVersions(key, appName, this.applicationFilter.branch)
+            .map((versions) => this.versions = versions);
+    };
+
```
sguiheux: We need that first élément to be empty, idem for branch. It allows user toi remove the filter

bnjjj: Yes but when you remove the filter it still takes the default branch in the filter. So there is always a filter if I can say that. Moreover, on the first visit when you don't choose filter branch or version it display an empty filter in the ui. I think it's better to not have multiple types in an array (string and number). And with this modification we don't have to override the dom with Jquery and it's better for angular bindings.

sguiheux: 1. In fact branch is mandatory, there is no sense to mix workflow from different branch.
Default branch filter has to be default branch from git repository
2. If you remove empty version, you must had something to remove the filter

bnjjj: The default version filter isn't the last version ? I mean, when your version filter is empty the version displayed is the last version ? Here the behavior is that when you don't select any filter it's empty. I don't understand why it is a problem to display the last version when you select a branch because it's the same behavior than before except that we have more information displayed and that give us the ability to kick jquery. I think there is a detail I don't catch. If you want I'm available on Gitter

bnjjj:
In fact I misunderstood, thanks for your time @sguiheux I think it's better now 🎉

PR #720, opened 2017-06-14
fix (worker): do not unregister worker to soon

fsamin:


engine/worker/cmd_main.go
```diff
@@ -253,6 +256,7 @@ func mainCommandRun(w *currentWorker) func(cmd *cobra.Command, args []string) {
 				}
 
 				// Unregister from engine
+				log.Debug("Jobs is done. Unregistering...")
```
yesnault: Job

PR #725, opened 2017-06-14
feat (cli): improve monitoring

yesnault:


cli/cds/ui/monitoring.go
```diff
@@ -544,6 +567,19 @@ func (ui *Termui) updateQueue(baseURL string) string {
...
+	switch status {
+	case sdk.StatusWaiting.String():
+		return "⏳ "
+	case sdk.StatusBuilding.String():
+		return "🔥 "
+	case sdk.StatusDisabled.String():
+		return "💀 "
+	case sdk.StatusChecking.String():
+		return "🔎 "
+	}
```
bnjjj: 🚨 or ❌ for fail ?

yesnault: There is no status fail on worker. Emoji are on worker status only (not on pipeline), but why not add it on pipeline above. I'll add another commit on this PR.

PR #766, opened 2017-06-26
feat (ui): new add step box

yesnault:


yesnault:
capture d ecran 2017-06-26 a 16 54 33

capture d ecran 2017-06-26 a 16 54 27

PR #792, opened 2017-06-30
feat (ui): add retry on webworker

sguiheux:


ui/src/assets/worker/web/log.js
```diff
@@ -21,21 +21,25 @@ onmessage = function (e) {
...
-        setInterval(function () {
-            var stepLogs = httpCall(url, api, user, session);
-            postMessage(stepLogs);
-            var jsonLogs = JSON.parse(stepLogs);
+
+        var xhr = httpCall(url, api, user, session);
+        if (xhr.status >= 400) {
+            return true;
+        }
+        if (xhr.status === 200 && xhr.responseText !== null) {
```
bnjjj: if xhr.responseText is undefined it will enter in the condition because you make a strict equals with null. Is it the right behavior ?

sguiheux: done

PR #843, opened 2017-07-12
fix(api): password variable name update is now enabled

bnjjj:


bnjjj:
@fsamin done
@sguiheux Already done for applications few weeks ago. I missed others, my bad

Issue #733, opened 2017-06-16
fix (ui): delete a hook


yesnault:
```
ERROR TypeError: Cannot read property 'name' of undefined
    at t.DZI8.t.removeHook (main.a08c423….bundle.js:1)
    at n.deleteHook (1.ddb4fdf….chunk.js:1)
    at Object.handleEvent (1.ddb4fdf….chunk.js:1)
    at Object.b [as handleEvent] (vendor.e275e1e….bundle.js:120)
    at Object.handleEvent (vendor.e275e1e….bundle.js:120)
    at Qt (vendor.e275e1e….bundle.js:71)
    at vendor.e275e1e….bundle.js:92
    at e.object.o [as _next] (vendor.e275e1e….bundle.js:176)
    at e.T14+.e.__tryOrUnsub (vendor.e275e1e….bundle.js:332)
    at e.T14+.e.next (vendor.e275e1e….bundle.js:332)
```

yesnault:
return this._applicationService.deleteHook(t.key, e.name, e.repositories_manager.name, n.id).map(function(n) {

yesnault:
+ check behaviour with delete app / pipeline

PR #848, opened 2017-07-12
fix (ui): delete a hook #733

bnjjj:


ui/src/app/views/application/show/workflow/tree/item/application.workflow.item.html
```diff
@@ -80,7 +80,7 @@
                         [scheduler]="s" [pipeline]="workflowItem.pipeline"></app-application-scheduler-item>
             </div>
             <div class="ui trigger segment" *ngFor="let h of workflowItem.hooks">
-                <app-application-hook-item [project]="project" [application]="workflowItem.application"
+                <app-application-hook-item [project]="project" [application]="application"
```
sguiheux: have to use workflowItem.application because of multi application workflow

PR #850, opened 2017-07-13
fix (api): fix oauth token creation for github

bnjjj:


fsamin:
something is missing somewhere; I suppose you run go this issue by using env variable or config file. Because it works fine with secret-backend like vault or the default one whiches are writing files :/

PR #853, opened 2017-07-13
fix (api): hook delete

bnjjj:


engine/api/main_routes.go
```diff
@@ -236,7 +236,7 @@ func (router *Router) init() {
 	router.Handle("/project/{key}/repositories_manager/{name}/application/{permApplicationName}/detach", POST(detachRepositoriesManager))
 	router.Handle("/project/{key}/application/{permApplicationName}/repositories_manager", GET(getRepositoriesManagerForApplicationsHandler))
 	router.Handle("/project/{key}/application/{permApplicationName}/repositories_manager/{name}/hook", POST(addHookOnRepositoriesManagerHandler))
-	router.Handle("/project/{key}/application/{permApplicationName}/repositories_manager/{name}/hook/{hookId}", DELETE(deleteHookOnRepositoriesManagerHandler))
+	router.Handle("/project/{key}/application/{permApplicationName}/repositories_manager/hook/{hookId}", DELETE(deleteHookOnRepositoriesManagerHandler))
```
fsamin: you have also to fix the sdk: https://github.com/ovh/cds/blob/master/sdk/repositories_manager.go#L254 and perhaps the cli

PR #872, opened 2017-07-20
fix (worker): stop register routine on single use. close #871

### LINKED ISSUE #871: --single-use option is not behaving as expected
Hi,

My worker is not behaving as expected.
I start my worker with: worker --api "https://xxx.xxx.xxx/cds/api" --token XXX --name buildbot-XXXXX_wheezy-amd64 --single-use

But my worker is not exiting after a build

fsamin:


fsamin:
Is something wrong with the build ?

PR #900, opened 2017-07-31
fix (api): persistent session token

fsamin:


engine/api/user/user.go
```diff
@@ -263,15 +283,95 @@ func NewPersistentSession(db gorp.SqlExecutor, u *sdk.User) (sessionstore.Sessio
...
+		case <-tick.C:
+			tx, err := DBFunc().Begin()
+			if err != nil {
+				continue
+			}
+
+			query := "select * from user_persistent_session where last_connection_date < $1"
+			tokens := []sdk.UserToken{}
+
+			if _, err := tx.Select(&tokens, query, time.Now().Add(-3*30*24*time.Hour)); err != nil {
```
bnjjj: a const for `-3*30*24*time.Hour` with a comment could be more readable

PR #926, opened 2017-08-06
fix (ui): }} in script editor #924

bnjjj:


bnjjj:
Good catch @sguiheux ! Now fixed :D thank you

PR #943, opened 2017-08-13
fix(ui): link github to project

bnjjj:
IMO there is a mistake here. But I'm not sure about that. In my personal computer it doesn't make sense to hide this when the name is github. But should we display this input code for bitbucket too ? 

Maybe @sguiheux could help me about that :) 


sguiheux:
Have to hide for github ans display for stash ans bitbucket. Maybe because if oauth version ?

yesnault:
yes, Github -> nothing to enter. Full tested, it's ok with https://github.com/ovh/cds/pull/946

PR #979, opened 2017-08-25
feat(action): add ovh-publish-metrics

d33d33:
This action push metrics to an OVH Metrics Data Platform account

### DIFF
--- a/contrib/actions/ovh-publish-metrics
+++ b/contrib/actions/ovh-publish-metrics
@@ -54,7 +54,7 @@ steps = [{
 
 set -e
 
-if [ "x{{name}}" != "x" ]; then
+if [ "{{.name}}" != "" ]; then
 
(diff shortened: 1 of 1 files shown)


contrib/actions/ovh-publish-metrics
```diff
@@ -0,0 +1,78 @@
...
+	 "labels" = {
+		type = "text"
+		description = "Labels of your metric (one K/V per line separated by a space)"
+		value = "app {{.cds.application}}
+env {{.cds.environment}}"
+	}
+	 "value" = {
+		type = "string"
+		description = "Value of your metric"
+		value = "T"
```
bnjjj: as a user I don't understand what means T. Maybe an example in description or better description ?

yesnault: or a link on metrics website on action description?

PR #997, opened 2017-08-30
Feat hatchery vsphere

bnjjj:
Don't merge


engine/hatchery/vsphere/client.go
```diff
@@ -0,0 +1,282 @@
...
+		log.Warning("createVMConfig> no network device found")
+		return nil, folder, fmt.Errorf("no network device found")
+	}
+
+	backing, errB := h.network.EthernetCardBackingInfo(ctx)
+	if errB != nil {
+		return nil, folder, sdk.WrapError(errB, "createVMConfig> cannot have ethernet backing info")
+	}
+
+	device, errE := object.EthernetCardTypes().CreateEthernetCard("e1000", backing)
```
sguiheux: e1000 this is the only choice or can be configurable? ( viper)

engine/hatchery/vsphere/spawn.go
```diff
@@ -0,0 +1,224 @@
...
+		"CDS_HATCHERY=" + fmt.Sprintf("%d", h.Hatchery().ID),
+		"CDS_HATCHERY_NAME=" + h.Hatchery().Name,
+		"CDS_BOOKED_JOB_ID=" + fmt.Sprintf("%d", jobID),
+		"CDS_TTL=" + fmt.Sprintf("%d", h.workerTTL),
+	}
+
+	env = append(env, getGraylogGrpcEnv(model)...)
+
+	script := fmt.Sprintf(
+		`cd $HOME; rm -f worker; curl "%s/download/worker/$(uname -m)" -o worker --retry 10 --retry-max-time 120 -C - >> /tmp/user_data 2>&1; chmod +x worker; ./worker`,
```
sguiheux: use value from viper: retry etc..

engine/hatchery/vsphere/spawn.go
```diff
@@ -0,0 +1,224 @@
...
+	if errR != nil {
+		return vm, sdk.WrapError(errR, "createVMModel> Cannot rename model %s", model.Name)
+	}
+
+	if _, err := task.WaitForResult(ctx, nil); err != nil {
+		return vm, sdk.WrapError(err, "createVMModel> error on waiting result for vm renaming %s", model.Name)
+	}
+
+	// FIXME doesn't work, give me a forbidden
+	return vm, h.client.WorkerModelUpdate(model.ID, model.Name, model.Type, model.Image, cdsclient.WorkerModelOpts.WithoutRegistrationNeed())
```
yesnault: this func create a "vm snapshot" for a worker model, but why does it need to update worker model? A worker model does not need to know if a "snapshot vm" exists or not.

engine/api/worker/model.go
```diff
@@ -50,7 +50,6 @@ func InsertWorkerModel(db gorp.SqlExecutor, model *sdk.Model) error {
 // UpdateWorkerModel update a worker model
 func UpdateWorkerModel(db gorp.SqlExecutor, model sdk.Model) error {
 	model.UserLastModified = time.Now()
-	model.NeedRegistration = true
```
yesnault: needRegistration is always true here. See review of 
```go
func (h *HatcheryVSphere) SpawnWorker
```
I'm not sure that all work on NeedRegistration is usefull.  See that irl :)

engine/hatchery/vsphere/spawn.go
bnjjj: Certainly a misunderstanding about flag `needRegistration`. In fact in my mind `needRegistration` is the flag that tell me the worker model was updated or need to create a new snapshot  in order to be update. So in this case my problem was to set `needRegistration` to false after creating a new snapshot/model. Am I wrong with this assessment ?

yesnault: needRegistration = ask worker model to run again "worker register", to let worker send all capabilities to API. In same func, Hatchery snaphosts the VM to make an image, but it's a specific use case of openstack & vsphere, because there is no flag "NeedRegistration Because User Update The Worker Model"
So, Hatchery OPenstack & VSphere does not need to update worker model when there create a image.

bnjjj: Ok right thanks ! So if I need to know if I have to create a new snapshot/model I must check the date and not this flag. :)

yesnault: yes. The openstack code of that:


```go
// if it's was a worker model for registration
// check if we need to create a new openstack image from it
// by comparing userDateLastModified from worker model
```

https://github.com/ovh/cds/blob/master/engine/hatchery/openstack/openstack.go#L196

sdk/hatchery/register.go
```diff
@@ -229,6 +229,7 @@ func workerRegister(h Interface, models []sdk.Model) error {
 			log.Warning("workerRegister> Too many errors on spawn with model %s, please check this worker model", m.Name)
 			continue
 		}
+		log.Debug("workerRegistrer> need registration %v", h.NeedRegistration(&m))
```
yesnault: two calls of h.NeedRegistration(&m) just for debug  -> add log in if statement below

PR #998, opened 2017-08-30
Feat fork pr

bnjjj:
Don't merge, the wip of fork and remote integration


fsamin:
few comments:
* what could we do for stash and gitlab ?
* didn't we need to update the gitclone actions ?
* had the remote as a tag in workflow run, could be interesting

PR #1008, opened 2017-09-01
fix (worker): env key

yesnault:
this avoid that ->
fork/exec /bin/sh : invalid argument
exec: not started

### DIFF
--- a/engine/hatchery/local/local.go
+++ b/engine/hatchery/local/local.go
@@ -138,7 +138,6 @@ func (h *HatcheryLocal) SpawnWorker(wm *sdk.Model, jobID int64, requirements []s
 		args = append(args, "register")
 	}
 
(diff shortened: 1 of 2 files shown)


fsamin:
could you explain ?

yesnault:
Value of private key are binary.

Put this variable in Command Environment gives:
```
fork/exec /bin/sh : invalid argument
```
with /bin/sh, zsh, bash... and at least on osx (and I don't know why exactly, it's too low level on os :) ).

Binary value in Environment variable is in any case not usable, so if private variable of type 'key' can be removed from environment var of script.
Public value (var name end with .pub) is oneline, no binary, it's ok to used it in step script.

Without that, worker running a step script, on a projet or app or env containing variable of type 'key' will always failed on osx (at least).

PR #1011, opened 2017-09-02
fix (plugin): sendLog

yesnault:

### DIFF
--- a/engine/worker/builtin.go
+++ b/engine/worker/builtin.go
@@ -103,7 +103,7 @@ func (w *currentWorker) runPlugin(ctx context.Context, a *sdk.Action, buildID in
 		}
 		for _, p := range params {
 			pluginArgs.Data[p.Name] = p.Value
-			if p.Type == sdk.SecretVariable {
+			if sdk.NeedPlaceholder(p.Type) {
 				pluginSecrets.Data[p.Name] = p.Value
 			}
 		}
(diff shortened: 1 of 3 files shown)


sdk/plugin/common.go
```diff
@@ -198,11 +199,18 @@ func SendLog(j IJob, format string, i ...interface{}) error {
...
 
 	s := fmt.Sprintf(format, i...)
+
+	for k, v := range j.Secrets().Data {
+		if len(v) >= 6 {
+			s = strings.Replace(s, v, "**"+k+"**", -1)
+		}
+	}
+
+	Trace.Println(s)
```
fsamin: \n has been removed ? are you sure ?

engine/worker/builtin.go
```diff
@@ -99,6 +103,9 @@ func (w *currentWorker) runPlugin(ctx context.Context, a *sdk.Action, buildID in
 		}
 		for _, p := range params {
 			pluginArgs.Data[p.Name] = p.Value
+			if p.Type == sdk.SecretVariable {
```
fsamin: not sure of this type. I think there is a func sdk.NeedPlaceHolder or something...

sdk/plugin/common.go
yesnault: Trace.Println(s) == Trace.Printf(s+"\n") :)

PR #1015, opened 2017-09-04
feat (action): action CDS_PythonPylint

ldec:
Hello,

I'll need to do a python pep8 check on multiple applications, I've written an action to do it properly with pylint. 

Louis

### DIFF
--- a/contrib/actions/cds-python-pylint.hcl
+++ b/contrib/actions/cds-python-pylint.hcl
@@ -1,0 +1,308 @@
+
+name = "CDS_PythonPylint"
+description = "Run pylint.
+By default, each pylint recommendation will be displayed as a Junit test.
+The short errors are fully contained inside the test title,
+an arrow (->) will be present if a part of the recommendation is displayed in
+the test body."
+
+// Requirements
+requirements = {
+    "python3.5" = {
+        type = "binary"
+        value = "python3.5"
+    },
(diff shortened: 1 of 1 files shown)


contrib/actions/cds-python-pylint.hcl
```diff
@@ -0,0 +1,306 @@
...
+        "extra_options": "{{.extra_options}}" or None
+    }
+
+    lint(**kwargs)
+
+EOF
+    },{
+        always_executed = true
+        JUnit = {
+            path = "{{.xml_output_file}}"
```
yesnault: it's 
```hcl
{
    always_executed = true
    jUnitReport = "{{.xml_output_file}}"
}
```

contrib/actions/cds-python-pylint.hcl
```diff
@@ -0,0 +1,306 @@
...
+description = "Run pylint.
+By default, each pylint recommendation will be displayed as a Junit test.
+The short errors are fully contained inside the test title,
+an arrow (->) will be present if a part of the recommendation is displayed in
+the test body."
+
+// Requirements
+requirements = {
+    "Python-3.5" = {
+        type = "model"
```
yesnault: It's not a best practice to force a model in Public Actions.

You should add requirement binary virtualenv and python3.5, but not model.

PR #1037, opened 2017-09-07
plugin-kafka-publisher: Allow reading kafka password from environment…

byo:
… variable.

Signed-off-by: Bartłomiej Święcki <bartlomiej.swiecki@corp.ovh.com>

### DIFF
--- a/contrib/plugins/plugin-kafka-publish/README.md
+++ b/contrib/plugins/plugin-kafka-publish/README.md
@@ -87,13 +87,13 @@ In a terminal
 In a terminal, go to the working directory in which you want to receive all CDS Data and run:
 
 ```shell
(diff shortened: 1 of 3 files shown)


byo:
@fsamin What's the current policy about backward-compatibility? If I convert this parameter to the named one, old usage will no longer be valid.

fsamin:
@byo, old binaries remain functionnal; and i prefer change usage now, for the best, and support the few users of this. 
Are you agree @yesnault @sguiheux @bnjjj ?

yesnault:
Yes, there is no good reason to put/keep password in arg. We have to remove it. A env var and (but not necessary), a user prompt should be better.

byo:
@fsamin Did update PR, please check.

PR #1040, opened 2017-09-08
fix (api): cds pipeline import existing stages

### LINKED ISSUE #923: Jobs not moved in the right stage during pipeline import
My jobs are not moved in the right stage.

Before the import the jobs are in separate stages.
The import tend to move every jobs in the same stage.
But the update works fine for the content of the jobs but not for the stage they're in.

yesnault:
remove all job before insert job from yml (if stage already exist)
use oldStage.ID to insertJob, this avoid create a new stage if the current stage already exists

close #923

### DIFF
--- a/engine/api/pipeline/pipeline_importer.go
+++ b/engine/api/pipeline/pipeline_importer.go
@@ -149,7 +149,7 @@ func ImportUpdate(db gorp.SqlExecutor, proj *sdk.Project, pip *sdk.Pipeline, msg
 				//Insert the job
 				j.PipelineStageID = oldStage.ID
 				j.Action.Type = sdk.JoinedAction
-				log.Debug(">> Creating job %s on stage %s on pipeline %s stageID:", j.Action.Name, s.Name, pip.Name, oldStage.ID)
+				log.Debug(">> Creating job %s on stage %s on pipeline %s stageID: %d", j.Action.Name, s.Name, pip.Name, oldStage.ID)
 				if err := InsertJob(db, j, oldStage.ID, pip); err != nil {
 					return sdk.WrapError(err, "ImportUpdate> Unable to insert job %s in %s", j.Action.Name, pip.Name)
 				}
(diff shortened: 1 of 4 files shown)


fsamin:
I think that the messages send through the channel will be confusing for the user. 
On an existing job, it will say "job xxx added, job yyy added" even if the first job already exists.
You should add a message "removing job xxx", then "job xxx added" or fix the update to avoid create new stage

yesnault:
update is fixed with this PR, avoiding create new stage is stage already exists.
+1 for msg: I'll update user msg in next commit on this PR.

yesnault:
@fsamin, like this? (it's in 4cfa2d2)
```bash
➜  cat sourcepipeline2.yml
name: mypipeline
stages:
  1|Stage 1:
    jobs:
      JobD:
        steps:
        - script: echo "jobD"

➜  cds -w pipeline import 00DD sourcepipeline2.yml --force
Updating stage Stage 1 ...
Deleting old jobs in stage Stage 1 ...
Job JobD in stage Stage 1 deleted
Inserting new jobs in stage Stage 1 ...
Job JobD in stage Stage 1 added
Stage Stage 1 updated
```

engine/api/pipeline/pipeline_importer.go
```diff
@@ -133,38 +135,26 @@ func ImportUpdate(db gorp.SqlExecutor, proj *sdk.Project, pip *sdk.Pipeline, msg
...
-						if msgChan != nil {
-							msgChan <- sdk.NewMessage(sdk.MsgPipelineJobUpdated, j.Action.Name, s.Name)
-						}
-						jobFound = true
-						break
-					}
+				//Insert the job
+				j.PipelineStageID = oldStage.ID
+				j.Action.Type = sdk.JoinedAction
+				log.Debug(">> Creating job %s on stage %s on pipeline %s stageID:", j.Action.Name, s.Name, pip.Name, oldStage.ID)
```
sguiheux: missing %d at the end

PR #1047, opened 2017-09-10
fix (api): goroutine on sanity

yesnault:
Sanity are too slow, really too slow on project with many apps / env.
Compute Warnings need to be refacto to be more precise, fastest without force user (with > 30 apps) to wait more dans 30" en each action on UI.
This commit in only to make UI faster, especially since the new UI get warnings async now.

### DIFF
--- a/engine/api/application_variable.go
+++ b/engine/api/application_variable.go
@@ -103,10 +103,11 @@ func restoreAuditHandler(w http.ResponseWriter, r *http.Request, db *gorp.DbMap,
 		return err
 	}
 
-	if err := sanity.CheckApplication(tx, p, app); err != nil {
-		log.Warning("restoreAuditHandler: Cannot check application sanity: %s\n", err)
-		return err
-	}
+	go func() {
+		if err := sanity.CheckApplication(tx, p, app); err != nil {
+			log.Warning("restoreAuditHandler: Cannot check application sanity: %s")
+		}
+	}()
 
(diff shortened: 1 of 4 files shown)


engine/api/application_variable.go
```diff
@@ -103,10 +103,11 @@ func restoreAuditHandler(w http.ResponseWriter, r *http.Request, db *gorp.DbMap,
 		return err
 	}
 
-	if err := sanity.CheckApplication(tx, p, app); err != nil {
-		log.Warning("restoreAuditHandler: Cannot check application sanity: %s\n", err)
-		return err
-	}
+	go func() {
+		if err := sanity.CheckApplication(tx, p, app); err != nil {
```
fsamin: don't put a transaction in a goroutine; you have to start a new one after commit the previous one.
you don't want your goroutine to rollback the whole tx if timeout excedeed

engine/api/application_variable.go
```diff
@@ -204,9 +205,11 @@ func deleteVariableFromApplicationHandler(w http.ResponseWriter, r *http.Request
 		return sdk.WrapError(err, "deleteVariableFromApplicationHandler: Cannot delete %s", varName)
 	}
 
-	if err := sanity.CheckApplication(tx, p, app); err != nil {
-		return sdk.WrapError(err, "restoreAuditHandler: Cannot check application sanity")
-	}
+	go func() {
+		if err := sanity.CheckApplication(tx, p, app); err != nil {
```
fsamin: start a new transaction

PR #1054, opened 2017-09-11
feat (hatchery): no need to check binary everytime

yesnault:
if requirement model/ network / hostname -> no need to check hatchery

Worker will be spawned and send spanInfos showing to users that binaries are not in path.

### DIFF
--- a/sdk/hatchery/hatchery.go
+++ b/sdk/hatchery/hatchery.go
@@ -211,6 +211,17 @@ func canRunJob(h Interface, timestamp int64, execGroups []sdk.Group, jobID int64
 		}
 	}
 
+	var containsModelRequirement, containsHostnameRequirement, containsNetworkRequirement bool
+	for _, r := range requirements {
+		switch r.Type {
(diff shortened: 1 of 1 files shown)


sdk/hatchery/hatchery.go
```diff
@@ -237,19 +248,27 @@ func canRunJob(h Interface, timestamp int64, execGroups []sdk.Group, jobID int64
...
-				if r.Value == c.Value || r.Value == c.Name {
-					found = true
-					break
+		// here, if there a model requirements, we don't need to check binaries
+		if r.Type == sdk.ModelRequirement && r.Value != model.Name {
+			log.Debug("canRunJob> %d - job %d - model requirement r.Value(%s) != model.Name(%s)", timestamp, jobID, r.Value, model.Name)
+			return false
+		}
+
+		if !containsModelRequirement && !containsHostnameRequirement && !containsNetworkRequirement {
```
fsamin: remove network requirement condition

PR #1071, opened 2017-09-13
feat (ui): more consistent menu

yesnault:
header project list / recent / search

for "normal user" (not admin), project list is really usefull
recently viewed and then search -> this encourages users to use the latest, and it's great (as more useful than search)


fsamin:
what about the project list with the "create project" link ?

PR #1116, opened 2017-09-19
 feat (api): implement Gitlab repository manager

proullon:


jhayotte:
Can't wait to have this PR merged!

engine/api/repositoriesmanager/repogitlab/client.go
```diff
@@ -0,0 +1,373 @@
...
+		URL:                   &url,
+		PushEvents:            &t,
+		MergeRequestsEvents:   &f,
+		TagPushEvents:         &f,
+		EnableSSLVerification: &f,
+	}
+
+	log.Debug("GitlabClient.CreateHook: %s %s\n", repo, *opt.URL)
+	_, _, err = c.client.Projects.AddProjectHook(repo, &opt)
+	if err != nil {
```
bnjjj: You can inline this

engine/api/repositoriesmanager/repogitlab/client.go
```diff
@@ -0,0 +1,373 @@
...
+		Context:     &cds,
+		State:       getGitlabStateFromStatus(eventpb.Status),
+		Ref:         &eventpb.BranchName,
+		TargetURL:   &url,
+		Description: &desc,
+	}
+
+	_, _, err := c.client.Commits.SetCommitStatus(eventpb.RepositoryFullname, eventpb.Hash, opt)
+	if err != nil {
+		return err
```
bnjjj: Inline

engine/api/repositoriesmanager/repogitlab/driver.go
```diff
@@ -0,0 +1,194 @@
...
+		if !ok {
+			return nil, fmt.Errorf("no app-id provided to Gitlab driver")
+		}
+		gd.ID = id
+
+		return gd, nil
+	}
+
+	err := json.Unmarshal([]byte(consumerData), &gd)
+	if err != nil {
```
bnjjj: Inline

engine/api/hook.go
```diff
@@ -21,6 +22,49 @@ import (
...
+
+	type gitlabEvent struct {
+		ObjectKind  string `json:"object_kind"`
+		Ref         string `json:"ref"`
+		UserName    string `json:"user_name"`
+		CheckoutSha string `json:"checkout_sha"`
+	}
+
+	var ge gitlabEvent
+	json.Unmarshal(data, &ge)
```
fsamin: check error and return if any

engine/api/repositoriesmanager/repogitlab/client.go
```diff
@@ -0,0 +1,373 @@
...
+}
+
+//PullRequestEvents is not implemented
+func (c *GitlabClient) PullRequestEvents(string, []interface{}) ([]sdk.VCSPullRequestEvent, error) {
+	return nil, fmt.Errorf("Not implemented on Gitlab")
+}
+
+// Release not implemented in Gitlab
+func (c *GitlabClient) Release(repo string, tagName string, title string, releaseNote string) (*sdk.VCSRelease, error) {
+	return nil, fmt.Errorf("Gitlab do not provide release system")
```
sguiheux: https://docs.gitlab.com/ee/api/tags.html#create-a-new-release

proullon: yeah I known haha, I'll change the error message, but I'll implement it later

proullon:
done

I'm not comfortable with inlined error though, it makes code harder to read since the function call is clogged with error management characters whereas the eye is used to ignore the usual error management:
```go
if err != nil {
   return err
}
```
Besides you can't always inline since you may have variables you want to keep out of error scope, so it's not possible to do this every time, then the code structure is eventually inconsistent...

Anyway, why do you want to inline errors ?

yesnault:
yes, we inline, when it's possible, the err management.
By inlining all existing err management, we discover so many errors not catched because of var err is used everywhere.

```
err:= aFunc()
if err != nil {
// err management
}

err:= anotherFunc()
//other code here, Oops, we forget manage err

err:= anotherFunc()
if err != nil {
// err management
}
```

By inlining, this avoid a lot of unchecked errors.

yesnault:
and a big thank you for this PR!

proullon:
I mean, errcheck ? https://github.com/kisielk/errcheck

yesnault:
why not, but I'm not sure that errcheck can detect  that anotherFunc2 is not check, because of err is used and check, after the call to anotherFunc3 (and this is the real issue: err affected, but not checked directly before being re-affected after).


```golang
err:= aFunc()
if err != nil {
// err management
}

err:= anotherFunc2()
//other code here, Oops, we forget manage err

err:= anotherFunc3()
if err != nil {
// err management
}
```

PR #1129, opened 2017-09-21
fix (worker): builtin gitclone / tag over ssh

yesnault:

### DIFF
--- a/sdk/vcs/git/git_tag.go
+++ b/sdk/vcs/git/git_tag.go
@@ -23,38 +23,53 @@ type TagOpts struct {
 
 // Tag makes git tag action
 func Tag(repo string, auth *AuthOpts, opts *TagOpts, output *OutputOpts) error {
-	if strings.HasPrefix(repo, "http://") || strings.HasPrefix(repo, "ftp://") || strings.HasPrefix(repo, "ftps://") {
-		return fmt.Errorf("Git protocol not supported")
-	}
-	if strings.HasPrefix(repo, "https://") {
-		return gitTagOverHTTPS(repo, auth, opts, output)
-	}
-	return gitTagOverSSH(auth, opts, output)
-}
-
-func gitTagOverHTTPS(repo string, auth *AuthOpts, opts *TagOpts, output *OutputOpts) error {
-	if auth == nil {
-		cmd := gitTagCommand(opts)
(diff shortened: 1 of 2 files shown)


sdk/vcs/git/git_tag.go
```diff
@@ -23,38 +23,53 @@ type TagOpts struct {
...
+		}
+		u.User = url.UserPassword(auth.Username, auth.Password)
+		return u.String(), nil
+	}
+	return repo, nil
+}
 
-	cmd := gitTagCommand(opts)
-	return runCommand(cmd, output)
+func gitTagOver(repo string, commands []cmd, auth *AuthOpts, output *OutputOpts) error {
```
fsamin: gitTagOver.. over what ?
It's so weird to pass the commands to this, i prefer as it was before

fsamin: Before, auth could be nil, now it returns an error. why ?

sdk/vcs/git/git_tag.go
```diff
@@ -23,38 +23,53 @@ type TagOpts struct {
...
+	if strings.HasPrefix(repo, "https://") {
+		return runCommand(commands, output)
+	}
+	return gitTagOverSSH(commands, auth, output)
 }
 
-func gitTagOverSSH(auth *AuthOpts, opts *TagOpts, output *OutputOpts) error {
+func gitTagOverSSH(commands []cmd, auth *AuthOpts, output *OutputOpts) error {
```
fsamin: gitTagOverSSH make a gittag over ssh... well, we doesn't have to pass commands in parameter

sdk/vcs/git/git_tag.go
```diff
@@ -74,10 +89,10 @@ func gitTagOverSSH(auth *AuthOpts, opts *TagOpts, output *OutputOpts) error {
 		return err
 	}
 
-	return runCommand(allCmd, output, "GIT_SSH="+wrapperPath)
+	return runCommand(commands, output, "GIT_SSH="+wrapperPath)
 }
 
-func gitTagCommand(opts *TagOpts) cmds {
+func gitTagCommand(repo string, opts *TagOpts) cmds {
```
fsamin: why did you add the repo, i don't see where you are using it ?

PR #1152, opened 2017-09-25
feat (hatchery/local) : provision workers

yesnault:


yesnault:
this will accelerate our it tests too

PR #1176, opened 2017-09-28
feat(api): add workflow_run status

bnjjj:


engine/sql/044_workflow_run_status.sql
```diff
@@ -0,0 +1,5 @@
+-- +migrate Up
+ALTER TABLE workflow_run ADD COLUMN status TEXT NOT NULL DEFAULT 'Building';
+
```
fsamin: update existing workflow_run

PR #1178, opened 2017-09-28
fix (worker, api, ui): Stop pipeline must kill worker running command…

### LINKED ISSUE #1173: On stop, worker should kill -9 its processes
Hi,

When asking a worker to stop, it would be great to kill every child process

Thanks

fsamin:
…. Closes #1173


bnjjj:
#1173

PR #1196, opened 2017-09-29
Dev/proullon/fix

proullon:


proullon:
still can't merge people ;)

yesnault:
@fsamin have to check before merging. We wait all approval from team (except on Holidays) before merging something :)

PR #1199, opened 2017-09-29
fix (api): no need to interpolate at this moment

yesnault:
all is already interpolate after

no infinite loop detected


yesnault:
The first run of of Workflow is OK too with this PR.
capture d ecran 2017-09-30 a 08 34 51

PR #1209, opened 2017-10-02
 fix (api): process workflow joins

fsamin:
* fix (api): process workflow joins
* feat (api): process workflow new attributes


sdk/workflow_run.go
```diff
@@ -22,6 +22,8 @@ type WorkflowRun struct {
 	WorkflowNodeRuns map[int64][]WorkflowNodeRun `json:"nodes" db:"-"`
 	Infos            []WorkflowRunInfo           `json:"infos" db:"-"`
 	Tags             []WorkflowRunTag            `json:"tags" db:"-"`
+	LastSubNumber    int64                       `json:"last_subnumber" db:"last_sub_num"`
```
bnjjj: why not `last_sub_num` as for db ? Or `last_sub_number` but change `num` to number for version number.

PR #1224, opened 2017-10-03
refactor(api): refactor stop workflow handlers

bnjjj:

### DIFF
--- a/engine/api/workflow_run.go
+++ b/engine/api/workflow_run.go
@@ -223,13 +223,16 @@ func (api *API) stopWorkflowRunHandler() Handler {
 
 		for _, wn := range run.WorkflowNodeRuns {
 			for _, wnr := range wn {
-				if wnr.SubNumber == run.LastSubNumber &&
-					(wnr.Status == sdk.StatusBuilding.String() || wnr.Status == sdk.StatusChecking.String() || wnr.Status == sdk.StatusWaiting.String()) {
-					if errS := workflow.StopWorkflowNodeRun(api.mustDB(), api.Cache, proj, wnr, stopInfos); errS != nil {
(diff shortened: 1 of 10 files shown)


engine/api/workflow_run.go
```diff
@@ -313,14 +311,19 @@ func (api *API) stopWorkflowNodeRunHandler() Handler {
...
 		if errC := tx.Commit(); errC != nil {
 			return sdk.WrapError(errC, "stopWorkflowNodeRunHandler> Unable to commit")
 		}
 
+		wr, errLw := workflow.LoadRun(api.mustDB(), key, name, number)
+		if errLw != nil {
+			return sdk.WrapError(errLw, "stopWorkflowNodeRunHandler> Unable to load workflow run %s", name)
+		}
+
+		if errR := workflow.ResyncWorkflowRunStatus(api.mustDB(), *wr); errR != nil {
```
fsamin: resync in the same transaction

engine/api/workflow_run.go
```diff
@@ -313,6 +311,11 @@ func (api *API) stopWorkflowNodeRunHandler() Handler {
 			return sdk.WrapError(errS, "stopWorkflowNodeRunHandler> Unable to stop workflow node run")
 		}
 
+		wr, errLw := workflow.LoadRun(tx, key, name, number)
+		if errLw != nil {
+			return sdk.WrapError(errLw, "stopWorkflowNodeRunHandler> Unable to load workflow run %s", name)
+		}
+
 		if errR := workflow.ResyncWorkflowRunStatus(tx, *wr); errR != nil {
```
fsamin: we should ResyncWorkflowRunStatus the pointer to the workflow run, so it could set the new status the the current workflow run object: 
https://github.com/ovh/cds/blob/refactor_stop_workflow/engine/api/workflow/resync_workflow.go#L59

PR #1255, opened 2017-10-06
fix (api): scheduler, sleep & db time

yesnault:

### DIFF
--- a/engine/api/scheduler/dao.go
+++ b/engine/api/scheduler/dao.go
@@ -39,8 +39,8 @@ func loadPipelineSchedulers(db gorp.SqlExecutor, query string, args ...interface
 
 func loadAndLockPipelineScheduler(db gorp.SqlExecutor, id int64) (*sdk.PipelineScheduler, error) {
 	query := `
-			SELECT 	pipeline_scheduler.* 
-			FROM 	pipeline_scheduler 
+			SELECT 	pipeline_scheduler.*
+			FROM 	pipeline_scheduler
 			WHERE   pipeline_scheduler.id = $1
 			FOR UPDATE NOWAIT`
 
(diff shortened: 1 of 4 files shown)


engine/api/scheduler/dao.go
```diff
@@ -233,7 +233,7 @@ func LoadPastExecutions(db gorp.SqlExecutor, id int64) ([]sdk.PipelineSchedulerE
 //LoadPendingExecutions loads all pipeline execution
 func LoadPendingExecutions(db gorp.SqlExecutor) ([]sdk.PipelineSchedulerExecution, error) {
 	as := []PipelineSchedulerExecution{}
-	if _, err := db.Select(&as, "select * from pipeline_scheduler_execution where executed = 'false' and execution_planned_date <=  $1", time.Now()); err != nil {
+	if _, err := db.Select(&as, "select * from pipeline_scheduler_execution where executed = 'false' and execution_planned_date <= now()"); err != nil {
```
fsamin: if you compare execution_planned_date with the database time, you should insert the database time

engine/api/scheduler/executer.go
```diff
@@ -76,6 +76,9 @@ func ExecuterRun(DBFunc func() *gorp.DbMap, store cache.Store) ([]sdk.PipelineSc
 			log.Warning("ExecuterRun> %s", err)
 			return nil, err
 		}
+
+		// wait, this avoid to re-run this job in another instance of API
+		time.Sleep(1 * time.Second)
```
fsamin: If for 1 loop, their is more than 60 pipelines to run; it will take 1 minutes to process all of them; and the ticker tick every 10 second, so it will burn :fire:

PR #1264, opened 2017-10-07
fix (api): load variable containing secret from worker

yesnault:
Keep value '{{.cds.foo}}' if foo is a app secret var to let worker use it

This use http://masterminds.github.io/sprig/defaults.html for default value if var is not replace with an existing variable

all funcs from http://masterminds.github.io/sprig/ are now available \o/


yesnault:
many secrets in value: replace default failed
missing UT
next commit will fix all

PR #1286, opened 2017-10-11
Change file download for x86_64 route

sfroment:
Since arch 386 isn't in TARGET_ARCH in the makefile it should be the amd64 file that is the default download
And x86_64 means 64bit processor so why download a 32bit binary since the automatic worker download is based on uname -m


fsamin:
You're absolutely right but, we also have to serve 32bits worker binary. So I would prefer, update the make file to build the right stuff and serve a new route `/download/worker/x86` for 32 bits binary and keep `/download/worker/x86_64` for 64 bits binary.

I raise issue #1287 in this way.

PR #1292, opened 2017-10-11
Add recursive option on GitClone

sfroment:


yesnault:
On an existing CDS Installation, GitClone won't be updated with new attribute. There no easy way to update builtin action on DB for now.

This PR needs a new migrate script under engine/sql/. This script have to update action_parameter table if GitClone action exist on table action.

Table action_edge_parameter does not need to be updated, default value for existing jobs will be empty string. But this will work only is default value of recursive is TRUE, even if there an error on ```strconv.ParseBool(recursive.Value)``` 

So, it's :

```golang
// recursive.Value could be empty on existing jobs
recursiveB, err := strconv.ParseBool(recursive.Value)
if err != nil {
recursiveB = true
}
```

Notice: we have planned an advanced mode on the attributes of the actions, this attribute will be under this mode.

yesnault:
there is a new feature on CDS: Advanced Parameter.
GitClone action contains two advanced parameter:
 - submodules (default true)
 - depth (default 50)

No parameter on CheckoutApplication, this action use the default parameters value.

There is more log on GitClone and CheckoutApplication to let user the git command launched by job.

It's closed with https://github.com/ovh/cds/pull/3171 (containing sql data migration).

PR #1293, opened 2017-10-11
fix (action): dockerPackage - don't rename attr

yesnault:


yesnault:
tests ok

PR #1343, opened 2017-10-20
feat (api): move gitlab in vcs µservice

sguiheux:


engine/vcs/types.go
```diff
@@ -84,6 +89,7 @@ var errGithubConfigurationError = fmt.Errorf("Github configuration Error")
 
 // GitlabServerConfiguration represents the gitlab configuration
 type GitlabServerConfiguration struct {
+	AppID  string `toml:"app-id" json:"-"`
```
yesnault: app-id or appID, no existing dash in configuration

PR #1346, opened 2017-10-22
feat (cli): add time/hostname/cdsname in status

yesnault:
This is useful to quick check if instances have not the same time

### DIFF
--- a/engine/api/event/event.go
+++ b/engine/api/event/event.go
@@ -71,10 +71,12 @@ func DequeueEvent(c context.Context) {
 	}
 }
 
+// GetHostname returns Hostname of this cds instance
 func GetHostname() string {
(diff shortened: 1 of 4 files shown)


engine/api/event/event.go
```diff
@@ -71,6 +71,14 @@ func DequeueEvent(c context.Context) {
 	}
 }
 
+func GetHostname() string {
```
sguiheux: doc

engine/api/event/event.go
```diff
@@ -71,6 +71,14 @@ func DequeueEvent(c context.Context) {
...
+	return hostname
+}
+
+func GetCDSName() string {
```
sguiheux: doc

PR #1369, opened 2017-10-24
feat (sdk,worker): add log fields CDSName / CDSVersion

yesnault:
example in graylog:

capture d ecran 2017-10-24 a 21 04 00


yesnault:
so, fields are added for all services using sdk.log, not only worker.

PR #1393, opened 2017-10-30
feat (engine): graylog on engine

yesnault:

### DIFF
--- a/engine/main.go
+++ b/engine/main.go
@@ -251,7 +251,16 @@ See $ engine config command for more details.
 		config()
 
 		//Initialize logs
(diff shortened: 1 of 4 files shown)


engine/main.go
```diff
@@ -280,49 +278,67 @@ See $ engine config command for more details.
 			signal.Stop(c)
 		}()
 
-		for _, a := range args {
-			var s Service
-			var cfg interface{}
+		type ServiceConf struct {
```
fsamin: what does it stand for ? why is it public ?

yesnault: public -> private
this struct is only here to collect args, to go start each service below. We start services, after init log (with name field with all names concat with '_')

PR #1404, opened 2017-10-31
fix (api): workflow name constraint

fsamin:


engine/sql/058_workflow_name_constraint.sql
```diff
@@ -0,0 +1,5 @@
+-- +migrate Up
+SELECT create_unique_index('workflow', 'IDX_WORKFLOW_NAME', 'name');
```
sguiheux: A workflow name must not be unique.
Couple project_id, worflow name is unique

PR #1417, opened 2017-11-03
feat (plugin/marathon): more logs

yesnault:


yesnault:
UT / IT / ITP OK

Issue #1416, opened 2017-11-03
fix (ui): display pipeline step logs


fsamin:
With a lot of logs when you click on a step to show the logs, it hangs for a while (no loading message), and it takes so much time to be displayed

PR #1418, opened 2017-11-03
fix(ui): display pipeline step logs #1416

bnjjj:


yesnault:
UT / IT / ITP OK

PR #1422, opened 2017-11-04
fix (api): more details on logs

yesnault:
Worker take a job, send info and there's errors on API (see below). This PR is only to add more context on logs.
Theses errors explains why a worker take a Job (workflow) and does not work on job after.

```
2017-11-04 15:44:20 | warn | POST    | 500  | /queue/workflows/8717/spawn/infos      postSpawnInfosWorkflowJobHandler> Cannot save job 8717: AddSpawnInfosNodeJobRun> Cannot update node job run: workflow.execute> Unable to execute node: Unable to update node id=2611 at status Building
2017-11-04 15:44:20 | err  | workflow.execute> Unable to add infos on run 554: loadRun> Unable to load workflow run%!(EXTRA string=select workflow_run.*
```


fsamin:
#1423

yesnault:
UT / IT / ITP OK

PR #1427, opened 2017-11-05
feat (sdk): remove new line on all log

yesnault:
this avoid having truncated log in graylog like:

```
| 2017-11-05 16:19:17 | err  | workflow.execute> Unable to add infos on run 588: loadRun> Unable to load workflow run. query:select workflow_run.*
```

this is better:
```
| 2017-11-05 17:20:56 | err  | workflow.execute> Unable to add infos on run 588: loadRun> Unable to load workflow run. query:select workflow_run.*     from workflow_run     where workflow_run.id = $1 for update nowait args:[588]: pq: current transaction is aborted, commands ignored until end of transaction block
```

### DIFF
--- a/sdk/log/log.go
+++ b/sdk/log/log.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"regexp"
 
 	log "github.com/Sirupsen/logrus"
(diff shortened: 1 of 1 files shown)


yesnault:
UT / IT / ITP OK

PR #1429, opened 2017-11-05
feat (api): workflow async handlers

### LINKED ISSUE #1423: feat (api): asynchronous routes for workflows
I see tons of logs on worker about timeout on API side; i think all PUT/POST operation on workflow node (sendind logs, variables and results) should be run asynchronously on  API.
I will modify the router to manage this through a redis queue without to much modification on the business code

fsamin:
to avoid timeout on worker side caused by node and jobs locks in database

close #1423


fsamin:
@yesnault fixed

yesnault:
UT / IT / ITP OK

PR #1430, opened 2017-11-05
fix (api): perf import pipeline

yesnault:
no need to call pipeline.LoadPipelines, it's already done inside sanity.CheckProjectPipelines


yesnault:
UT / IT / ITP OK

PR #1432, opened 2017-11-06
fix (sdk): do not use pointer on Error

yesnault:
it was the cause of "forbidden (caused by: unknown worker 'fooooooo' Name:foo GroupID:000 ModelID:000 Model: HatcheryID:000 HatcheryName:foo)" on a 403 for a user.


yesnault:
UT / IT OK

PR #1445, opened 2017-11-08
Feat workflow collector

bnjjj:
Please CHECK SQL MIGRATION NUMBER before merge :)

### DIFF
--- a/engine/api/worker/worker.go
+++ b/engine/api/worker/worker.go
@@ -68,17 +68,17 @@ func DeleteWorker(db *gorp.DbMap, id string) error {
 		switch jobType.String {
 		case sdk.JobTypePipeline:
 			if err := pipeline.RestartPipelineBuildJob(tx, jobID.Int64); err != nil {
-				log.Error("DeleteWorker[%d]> Cannot restart pipeline build job: %s", id, err)
+				log.Error("DeleteWorker[%s]> Cannot restart pipeline build job: %s", name, err)
 			} else {
-				log.Info("DeleteWorker[%d]> PipelineBuildJob %d restarted after crash", id, jobID.Int64)
+				log.Info("DeleteWorker[%s]> PipelineBuildJob %d restarted after crash", name, jobID.Int64)
 			}
 		case sdk.JobTypeWorkflowNode:
 			wNodeJob, errL := workflow.LoadNodeJobRun(tx, nil, jobID.Int64)
 			if errL == nil && wNodeJob.Retry < 3 {
 				if err := workflow.RestartWorkflowNodeJob(db, *wNodeJob); err != nil {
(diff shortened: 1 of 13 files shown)


engine/api/worker/worker.go
```diff
@@ -44,27 +45,45 @@ func DeleteWorker(db *gorp.DbMap, id string) error {
...
+
+		switch pbJobType.String {
+		case sdk.JobTypePipeline:
+			if err := pipeline.RestartPipelineBuildJob(tx, jobID.Int64); err != nil {
+				log.Error("DeleteWorker[%d]> Cannot restart pipeline build job: %s", id, err)
+			} else {
+				log.Info("DeleteWorker[%d]> PipelineBuildJob %d restarted after crash", id, jobID.Int64)
+			}
+		case sdk.JobTypeWorkflowNode:
+			wNodeJob, errL := workflow.LoadNodeJobRunWithoutStore(tx, jobID.Int64)
```
fsamin: why not modify LoadNodeJobRun to handle nil store ?

engine/api/workflow/dao_node_run_job.go
```diff
@@ -101,6 +101,17 @@ func LoadNodeJobRun(db gorp.SqlExecutor, store cache.Store, id int64) (*sdk.Work
 	return &job, nil
 }
 
+//LoadNodeJobRunWithoutStore load a NodeJobRun given its ID without a given store
+func LoadNodeJobRunWithoutStore(db gorp.SqlExecutor, id int64) (*sdk.WorkflowNodeJobRun, error) {
```
fsamin: i think it better to only keep LoadNodeJobRun

engine/sql/061_pipeline_build_job_retry.sql
```diff
@@ -0,0 +1,6 @@
+-- +migrate Up
+ALTER TABLE workflow_node_run_job ADD COLUMN retry int DEFAULT 0;
+ALTER TABLE worker ADD COLUMN job_type text;
```
fsamin: should we update workflow_node_run_job to set the default value on existing tuples ?

bnjjj: Yes we can !

bnjjj:
In order to prod you need to stop hatchery, wait all workers are gone and prod

engine/sql/061_pipeline_build_job_retry.sql
bnjjj: No, see PR comment below

yesnault:
rebased master & resolving conflicts done

yesnault:
Ready to merge, run 629. UT / IT OK

PR #1464, opened 2017-11-10
feat (api, ui): move run condition from trigger to node

sguiheux:
SQL 63


engine/api/workflow/dao_node.go
```diff
@@ -156,6 +157,12 @@ func insertNodeContext(db gorp.SqlExecutor, c *sdk.WorkflowNodeContext) error {
 		sqlContext.DefaultPipelineParameters = sql.NullString{String: string(b), Valid: true}
 	}
 
+	cond, errC := json.Marshal(c.Conditions)
```
fsamin: you can use `sqlContext.Conditions, err = gorpmapping.JSONToNullString(c.Conditions)`

engine/api/workflow/dao_node.go
```diff
@@ -273,6 +280,12 @@ func loadNodeContext(db gorp.SqlExecutor, store cache.Store, wn *sdk.WorkflowNod
 		}
 		ctx.Environment = env
 	}
+
+	if sqlContext.Conditions.Valid {
```
fsamin: you can use `gorpmapping.JSONNullString(sqlContext.Conditions, &ctx.Conditions)`

PR #1468, opened 2017-11-12
fix (api): more precision in the log

yesnault:
There is something strange,

```
Run> Unable to load to pipeline scheduler 150 pq: could not obtain lock on row in relation "pipeline_scheduler"
```

This error should be "55P03", so ignored from log.Error. This commit is to have the real pqerr.Code

### DIFF
--- a/engine/api/error.go
+++ b/engine/api/error.go
@@ -15,7 +15,7 @@ func WriteError(w http.ResponseWriter, r *http.Request, err error) {
 
 	// ErrAlreadyTaken is not useful to log in warning
 	if errProcessed.ID == sdk.ErrAlreadyTaken.ID {
-		log.Debug("%-7s | %-4d | %s \t %s", r.Method, errProcessed.Status, r.RequestURI, err)
+		log.Info("%-7s | %-4d | %s \t %s", r.Method, errProcessed.Status, r.RequestURI, err)
 	} else {
 		log.Warning("%-7s | %-4d | %s \t %s", r.Method, errProcessed.Status, r.RequestURI, err)
 	}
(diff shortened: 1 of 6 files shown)


engine/api/error.go
```diff
@@ -10,8 +10,15 @@ import (
...
 func WriteError(w http.ResponseWriter, r *http.Request, err error) {
 	al := r.Header.Get("Accept-Language")
-	msg, code := sdk.ProcessError(err, al)
+	msg, errProcessed := sdk.ProcessError(err, al)
 	sdkErr := sdk.Error{Message: msg}
-	log.Warning("%-7s | %-4d | %s \t %s", r.Method, code, r.RequestURI, err)
-	WriteJSON(w, r, sdkErr, code)
+
+	// ErrAlreadyTaken is not useful to log in warning
+	if errProcessed.ID == sdk.ErrAlreadyTaken.ID {
```
fsamin: sdk.ErrorIs ?

engine/api/error.go
```diff
@@ -10,8 +10,15 @@ import (
...
+		log.Debug("%-7s | %-4d | %s \t %s", r.Method, errProcessed.Status, r.RequestURI, err)
+	} else {
+		log.Warning("%-7s | %-4d | %s \t %s", r.Method, errProcessed.Status, r.RequestURI, err)
```
fsamin: I find it really interesting to see that there is a lot of contention on every jobs, It means we still have things to improve.
If you are upset with the warning, the log .Info would be fine for me if we are currently using the info log-level

yesnault: yes, Info

engine/api/error.go
yesnault: ? (I don't understand the comment :) )

yesnault: hum, ok. errorIs(...), I'll fix

PR #1491, opened 2017-11-14
fix (api): speedup attachPipelinesToApplication

yesnault:


yesnault:
it's for you @henyxia

PR #1506, opened 2017-11-17
fix(api): fix workflow manual run

bnjjj:


bnjjj:
Nope it's good, check run 670 (rebased with master)

PR #1519, opened 2017-11-19
feat (api): allow listen on an addr only (ex:127.0.0.1)

yesnault:

### DIFF
--- a/engine/api/api.go
+++ b/engine/api/api.go
@@ -45,7 +45,7 @@ type Configuration struct {
 		UI  string `toml:"ui" default:"http://localhost:2015"`
 	} `toml:"url" comment:"#####################\n CDS URLs Settings \n####################"`
 	HTTP struct {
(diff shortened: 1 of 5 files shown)


engine/api/api.go
```diff
@@ -45,8 +45,9 @@ type Configuration struct {
 		UI  string `toml:"ui" default:"http://localhost:2015"`
 	} `toml:"url" comment:"#####################\n CDS URLs Settings \n####################"`
 	HTTP struct {
-		Port       int `toml:"port" default:"8081"`
-		SessionTTL int `toml:"sessionTTL" default:"60"`
+		Addr       string `toml:"addr" default:"127.0.0.1" commented:"true" comment:"Listen address without port"`
```
fsamin: humm not sure it's good idea to set by default 127.0.0.1, i won't be enough with the docker-compose file; 0.0.0.0 (= all interfaces) should work

PR #1540, opened 2017-11-21
fix (api): quick check status on take

yesnault:
check status without FOR UPDATE
then if status is not waiting, select again with FOR UPDATE

this will avoid FOR UPDATE select for many existing workers who try to take the job

postTakeWorkflowJobHandler> Cannot take job 12043: TakeNodeJobRun> job 12043 is not waiting status. Current status:Building: This job is already taken by another worker

### DIFF
--- a/engine/api/workflow/execute_node_job_run.go
+++ b/engine/api/workflow/execute_node_job_run.go
@@ -128,10 +128,9 @@ func prepareSpawnInfos(j *sdk.WorkflowNodeJobRun, infos []sdk.SpawnInfo) error {
 // TakeNodeJobRun Take an a job run for update
 func TakeNodeJobRun(db gorp.SqlExecutor, store cache.Store, p *sdk.Project, jobID int64, workerModel string, workerName string, workerID string, infos []sdk.SpawnInfo, chanEvent chan<- interface{}) (*sdk.WorkflowNodeJobRun, error) {
 	// first load without FOR UPDATE WAIT to quick check status
-	query := `SELECT status FROM workflow_node_run_job WHERE id = $1`
-	var currentStatus string
-	if err := db.QueryRow(query, jobID).Scan(&currentStatus); err != nil {
-		return nil, sdk.WrapError(err, "workflow.UpdateNodeJobRunStatus> Cannot select status from workflow_node_run_job node job run %d", jobID)
+	currentStatus, errS := db.SelectStr(`SELECT status FROM workflow_node_run_job WHERE id = $1`, jobID)
+	if errS != nil {
+		return nil, sdk.WrapError(errS, "workflow.UpdateNodeJobRunStatus> Cannot select status from workflow_node_run_job node job run %d", jobID)
 	}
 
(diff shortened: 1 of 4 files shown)


yesnault:
PR TBD

engine/api/workflow/execute_node_job_run.go
```diff
@@ -123,20 +123,29 @@ func prepareSpawnInfos(j *sdk.WorkflowNodeJobRun, infos []sdk.SpawnInfo) error {
 
 // TakeNodeJobRun Take an a job run for update
 func TakeNodeJobRun(db gorp.SqlExecutor, store cache.Store, p *sdk.Project, id int64, workerModel string, workerName string, workerID string, infos []sdk.SpawnInfo, chanEvent chan<- interface{}) (*sdk.WorkflowNodeJobRun, error) {
-	job, err := LoadAndLockNodeJobRunWait(db, store, id)
+	// first load without FOR UPDATE WAIT to quick check status
```
fsamin: if we want to check the status, maybe we should only make a precise query to get only the node status without all the node structure

engine/api/workflow/execute_node_job_run.go
```diff
@@ -95,16 +95,16 @@ func UpdateNodeJobRunStatus(db gorp.SqlExecutor, store cache.Store, p *sdk.Proje
 
 // AddSpawnInfosNodeJobRun saves spawn info before starting worker
 func AddSpawnInfosNodeJobRun(db gorp.SqlExecutor, store cache.Store, p *sdk.Project, id int64, infos []sdk.SpawnInfo) (*sdk.WorkflowNodeJobRun, error) {
-	j, err := LoadAndLockNodeJobRunWait(db, store, id)
+	j, err := LoadAndLockNodeJobRunNoWait(db, store, id)
```
fsamin: should we raise the number of retry on the asynchronous handler ?

engine/api/workflow/execute_node_job_run.go
```diff
@@ -145,17 +154,29 @@ func TakeNodeJobRun(db gorp.SqlExecutor, store cache.Store, p *sdk.Project, id i
...
 	if err := UpdateNodeJobRunStatus(db, store, p, job, sdk.StatusBuilding, chanEvent); err != nil {
 		log.Debug("TakeNodeJobRun> call UpdateNodeJobRunStatus on job %d set status from %s to %s", job.ID, job.Status, sdk.StatusBuilding)
-		return nil, sdk.WrapError(err, "TakeNodeJobRun>Cannot update node job run")
+		return nil, sdk.WrapError(err, "TakeNodeJobRun>Cannot update node job run %d", id)
 	}
 
 	return job, nil
 }
 
+func checkStatusWaiting(store cache.Store, id int64, job *sdk.WorkflowNodeJobRun) error {
```
fsamin: if we only need the nodejobrun status, don't pass the whole object

fsamin: rename id => jobID

PR #1543, opened 2017-11-22
feat (api): export application with encrypted secrets

fsamin:


engine/api/project/dao_key.go
```diff
@@ -56,3 +56,22 @@ func DeleteProjectKey(db gorp.SqlExecutor, projectID int64, keyName string) erro
 	_, err := db.Exec("DELETE FROM project_key WHERE project_id = $1 AND name = $2", projectID, keyName)
 	return sdk.WrapError(err, "DeleteProjectKey> Cannot delete key %s", keyName)
 }
+
+func loadBuildinKey(db gorp.SqlExecutor, projectID int64) (k sdk.ProjectKey, err error) {
```
bnjjj: why named return ?

sdk/exportentities/application.go
```diff
@@ -69,242 +33,54 @@ func NewApplication(app *sdk.Application) (a *Application) {
 
 	a.Variables = make(map[string]VariableValue, len(app.Variable))
 	for _, v := range app.Variable {
+		fmt.Println(v)
```
bnjjj: to delete

engine/api/project/dao_key.go
fsamin: All return statements precise the returned values
Here it's usefull to init the returned values which are used everywhere

bnjjj: ok not really fan but ok, however at this line https://github.com/ovh/cds/pull/1543/files/690a39286457e62bcf846a7b73c1706fbe5a4e56#diff-7f65e53ab8b93430efdc845089d28284R62 you make some kind of variable shadowing

PR #1545, opened 2017-11-22
feat (api): fix add tag / resync / node name

sguiheux:


sdk/workflow.go
```diff
@@ -92,6 +92,23 @@ func (w *Workflow) Nodes() []int64 {
...
+	}
+	for _, j := range w.Joins {
+		for _, t := range j.Triggers {
+			n = t.WorkflowDestNode.GetNodeByName(name)
+			if n != nil {
+				return n
+			}
+		}
+	}
+	return nil
```
fsamin: it could be nice to return a (non blocking ?) error to tell the user that all have not been resynchronized

PR #1547, opened 2017-11-22
feat (api): pprof handlers

fsamin:
The API handles /admin/debug for admins. You just have to add your session token as a query param.

capture d ecran 2017-11-22 a 21 59 00

It is inspired by the net/http/pprof package, with a few custom things.
I hope we will be able to embbed the next pprof web ui from go 1.10 https://twitter.com/rakyll/status/887401222478602241?lang=fr

### DIFF
--- a/engine/api/api.go
+++ b/engine/api/api.go
@@ -250,6 +250,18 @@ func (a *API) CheckConfiguration(config interface{}) error {
 	return nil
 }
 
+func getUserSession(c context.Context) string {
+	i := c.Value(auth.ContextUserSession)
+	if i == nil {
+		return ""
+	}
+	u, ok := i.(string)
+	if !ok {
+		return ""
+	}
+	return u
+}
+
 func getUser(c context.Context) *sdk.User {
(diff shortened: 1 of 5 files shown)


yesnault:
I'm not fan of query params for session-t (GET). It could be catch by logger on proxy.
Could we add all three commands in cds ?

```
cds admin debug stack: debug/goroutine?debug=2"> Full goroutine stack dump
cds admin debug trace 5: debug/trace?seconds=5"> Trace profile for 5 seconds
cds admin debug cpu 5: debug/cpu?seconds=5"> CPU profile for 5 seconds
```

fsamin:
True, but we can log headers on a reverse proxy too (example: http://httpd.apache.org/docs/2.2/mod/mod_log_config.html).

Obviously we still don't need it through we webui reverse proxy.

A cookie would solve it nicelier too...

Indeed request params, headers and cookies can be catched by reverse proxies.... as well the reverse proxy can catch the password of all your users :trollface: 

Well a curl is enough to get the data, we don't have to implement new commands in the cli.

PR #1595, opened 2017-11-28
feat (api): import application

fsamin:


engine/api/application_import.go
```diff
@@ -0,0 +1,163 @@
...
+		case "application/json":
+			errapp = json.Unmarshal(body, eapp)
+		case "application/x-yaml", "text/x-yam":
+			errapp = yaml.Unmarshal(body, eapp)
+		default:
+			return sdk.NewError(sdk.ErrWrongRequest, fmt.Errorf("unsupported content-type: %s", contentType))
+		}
+
+		if errapp != nil {
+			return sdk.NewError(sdk.ErrWrongRequest, err)
```
bnjjj: s/err/errapp

engine/api/application_import.go
```diff
@@ -0,0 +1,163 @@
...
+		}
+
+		log.Info("postApplicationImportHandler> Import application %s in project %s (force=%v)", eapp.Name, key, force)
+		log.Debug("postApplicationImportHandler> App: %+v", eapp)
+
+		//Check valid application name
+		regexp := regexp.MustCompile(sdk.NamePattern)
```
bnjjj: don't name it regexp like the package, it's confusing

engine/api/application_import.go
```diff
@@ -0,0 +1,163 @@
...
+
+		tx, errtx := api.mustDB().Begin()
+		if errtx != nil {
+			return sdk.WrapError(errtx, "postApplicationImportHandler> Unable to start transaction")
+
+		}
+
+		defer tx.Rollback()
+
+		done := make(chan bool)
```
bnjjj: I have read that a best practice when we use a chan to inform that we're done it's better to use `chan struct{}` less memory space and more explicit that we don't care about the value

fsamin: it's copy & paste from old code, i will change it for a waitgroup

PR #1608, opened 2017-11-30
feat (api): import application with secret values

fsamin:


engine/api/application/application_importer_pipeline.go
```diff
@@ -0,0 +1,143 @@
+package application
+
+import (
+	"github.com/go-gorp/gorp"
+	"github.com/ovh/cds/engine/api/cache"
```
sguiheux: new line

engine/api/keys/keys_pgp.go
```diff
@@ -0,0 +1,124 @@
...
+		return nil, sdk.WrapError(err, "GeneratePGPPublicKey> Cannot serialize public key")
+	}
+	return bufPublic, nil
+}
+
+// GeneratePGPKeyPair generates a private / public PGP key
+func GeneratePGPKeyPair(name string) (id string, pub string, priv string, err error) {
+	key, err := NewOpenPGPEntity(name)
+	if err != nil {
+		return
```
bnjjj: Proposal: in case of named return, can we still write the err variable after the return ? In order to directly see that we return values and which. And not each time go above to see the function declaration

PR #1611, opened 2017-11-30
fix (api): 204 for no content

yesnault:
In a handler, we can't return nil anymore.

a http code != 204, and < 300 -> Angular 5 will always check body
For all handlers with no-content (and returns OK), we have to return a 204 only.

This pr fix: delete worker model, create group, update group (add / remove user), delete group

Relevant code on angular 5 about that:
https://github.com/angular/angular/blob/65a2cb830716afaace03684e4023e7b0e24d669f/packages/common/http/src/xhr.ts#L165


bnjjj:
interesting :)

PR #1613, opened 2017-12-01
Xmpp html

landru29:
Send XMPP message as HTML to highlight notifications.
ref: https://github.com/mattn/go-xmpp/blob/master/xmpp.go#L644


fsamin:
doc/tutorials/workers.md ?

PR #1614, opened 2017-12-01
fix (api): 204 for no content

yesnault:
In a handler, we can't return nil anymore.

a http code != 204, and < 300 -> Angular 5 will always check body
For all handlers with no-content (and returns OK), we have to return a 204 only.

This pr fix: delete worker model, create group, update group (add / remove user), delete group

Relevant code on angular 5 about that:
https://github.com/angular/angular/blob/65a2cb830716afaace03684e4023e7b0e24d669f/packages/common/http/src/xhr.ts#L165

### DIFF
--- a/engine/api/api_routes.go
+++ b/engine/api/api_routes.go
@@ -12,6 +12,7 @@ func (api *API) InitRouter() {
 	api.Router.SetHeaderFunc = DefaultHeaders
 	api.Router.Middlewares = append(api.Router.Middlewares, api.authMiddleware)
 	api.Router.PostMiddlewares = append(api.Router.PostMiddlewares, api.deletePermissionMiddleware)
+	api.Router.PostMiddlewares = append(api.Router.PostMiddlewares, api.writeNoContentPostMiddleware)
 	api.lastUpdateBroker = &lastUpdateBroker{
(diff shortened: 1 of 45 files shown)


engine/api/api_routes.go
```diff
@@ -12,6 +12,7 @@ func (api *API) InitRouter() {
 	api.Router.SetHeaderFunc = DefaultHeaders
 	api.Router.Middlewares = append(api.Router.Middlewares, api.authMiddleware)
 	api.Router.PostMiddlewares = append(api.Router.PostMiddlewares, api.deletePermissionMiddleware)
+	api.Router.PostMiddlewares = append(api.Router.PostMiddlewares, api.writeNoContentPostMiddleware)
```
fsamin: add it in the line above 
```
api.Router.PostMiddlewares = append(api.Router.PostMiddlewares, api.deletePermissionMiddleware, api.writeNoContentPostMiddleware)
```

fsamin: Why a middleware ? you don't want to use it in other µServices ?

yesnault:
capture d ecran 2017-12-01 a 18 21 17

PR #1617, opened 2017-12-01
refactor(api): Refactor stop workflow run

bnjjj:


bnjjj:
I made a mistake and push bad branch, this one is already merged, I don't understand why github show us the same diff but ... My branch is about the start workflow run and not the stop. It's weird :o

PR #1618, opened 2017-12-01
Remove reference to the `parallel` option in the venom plugin.

sean-:
Simple doc cleanup.  Also added the `vars-from-file` while here.


yesnault:
hum, readme have to be updated with vendor too.

PR #1619, opened 2017-12-02
feat(ui): add better diff view for pipeline audit

bnjjj:

capture d ecran 2017-12-02 a 12 09 18


yesnault:
Can you add something on the table above to see on which line we are ?

bnjjj:
done @yesnault and updated with componentFactory @sguiheux

PR #1623, opened 2017-12-04
Refactor start workflow node run

bnjjj:


engine/api/workflow_run.go
```diff
@@ -497,34 +507,33 @@ func startWorkflowRun(chEvent chan<- interface{}, chError chan<- error, db *gorp
...
-			if lastRun != nil {
-				var errmr error
-				_, errmr = workflow.ManualRunFromNode(tx, store, p, wf, lastRun.Number, opts.Manual, fromNode.ID, chEvent)
-				if errmr != nil {
-					chError <- sdk.WrapError(errmr, "postWorkflowRunHandler> Unable to run workflow from node")
+		for i := 0; i < len(fromNodes); i++ {
+			select {
+			case <-workerOptions.chanNodeRun:
+			case err := <-workerOptions.chanError:
+				if chError != nil {
```
fsamin: if chanError is nil, log the error

engine/api/workflow_run.go
```diff
@@ -541,6 +550,59 @@ func startWorkflowRun(chEvent chan<- interface{}, chError chan<- error, db *gorp
 	}
 }
 
+func runFromNodeWorker(db *gorp.DbMap, store cache.Store, opts sdk.WorkflowRunPostHandlerOption, p *sdk.Project, wf *sdk.Workflow, lastRun *sdk.WorkflowRun, u *sdk.User, workerOptions *workerOpts) {
```
fsamin: i would prefer to not use the word `worker` here to avoid confusions

PR #1636, opened 2017-12-04
feat (worker): worker requirements command

### LINKED ISSUE #1604: New worker command to get requirement services and their variables
Ideas of options:
- json output could be cool
- shell output can be evaled to retrieve contents as shell variables

yesnault:
close #1604

### DIFF
--- a/engine/worker/cmd_requirements.go
+++ b/engine/worker/cmd_requirements.go
@@ -1,0 +1,101 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"os"
+	"strconv"
+	"time"
+
+	"github.com/spf13/cobra"
+
+	"github.com/ovh/cds/sdk"
+)
+
+var cmdRequirementsFormat string
+
+func cmdRequirements(w *currentWorker) *cobra.Command {
+	c := &cobra.Command{
+		Use:   "requirements",
+		Short: "worker requirements --format json|csv|keyval",
+		Run:   requirementsCmd(w),
+	}
+	c.Flags().StringVar(&cmdRequirementsFormat, "format", "json", "Output format. json, csv, keyval")
+	return c
+}
+
+func requirementsCmd(w *currentWorker) func(cmd *cobra.Command, args []string) {
+	return func(cmd *cobra.Command, args []string) {
+		portS := os.Getenv(WorkerServerPort)
+		if portS == "" {
+			sdk.Exit("%s not found, are you running inside a CDS worker job?\n", WorkerServerPort)
+		}
+
+		port, errPort := strconv.Atoi(portS)
+		if errPort != nil {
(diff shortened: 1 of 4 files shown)


engine/worker/cmd_requirements.go
```diff
@@ -0,0 +1,101 @@
...
+
+	"github.com/ovh/cds/sdk"
+)
+
+var cmdRequirementsFormat string
+
+func cmdRequirements(w *currentWorker) *cobra.Command {
+	c := &cobra.Command{
+		Use:   "requirements",
+		Short: "worker requirements --format json|csv|keyval",
```
fsamin: i would prefer `env` over `keyval`and output `export KEY=val`so we use it as :
```
eval $(worker requirements --format env)
```

engine/worker/cmd_requirements.go
```diff
@@ -0,0 +1,101 @@
...
+			sdk.Exit("%s not found, are you running inside a CDS worker job?\n", WorkerServerPort)
+		}
+
+		port, errPort := strconv.Atoi(portS)
+		if errPort != nil {
+			sdk.Exit("cannot parse '%s' as a port number", portS)
+		}
+
+		req, errRequest := http.NewRequest("GET", fmt.Sprintf("http://127.0.0.1:%d/requirements", port), nil)
+		req.Header.Add("Content-Type", "application/json")
```
fsamin: You have no body, so you don't need content-type

engine/worker/cmd_requirements.go
```diff
@@ -0,0 +1,101 @@
...
+
+		var reqs []sdk.Requirement
+		if err := json.Unmarshal(body, &reqs); err != nil {
+			sdk.Exit("Get Requirements failed: unable to unmarshal body %v\n", err)
+		}
+
+		if resp.StatusCode < 300 {
+			switch cmdRequirementsFormat {
+			case "json":
+				fmt.Printf("body: %+v\n", string(body))
```
fsamin: Why not just`fmt.Println(string(body))` so we would use it as follow:
`worker requirements --format json | jq '.[]' ` or `worker requirements --format json > req.json`

engine/worker/cmd_requirements.go
```diff
@@ -0,0 +1,101 @@
...
+			return
+		}
+		cdsError := sdk.DecodeError(body)
+		sdk.Exit("Get Requirements failed: %v\n", cdsError)
+	}
+}
+
+func printCSVRequirements(requirements []sdk.Requirement) {
+	for _, req := range requirements {
+		fmt.Printf("%s;%s;%s\n", req.Type, req.Name, req.Value)
```
fsamin: we should quote value, to avoid problem with the complex services value

sguiheux:
Can I have an example of how to extract pg hostname for a service ? on a job with a pg and a redis for example
I think it will be too complex.

yesnault:
Another PR on this subject will be done, using {{.cds.requirements...}}

PR #1637, opened 2017-12-05
refactor (api): move parse and import business app code to package

fsamin:


engine/api/application/application_parser.go
```diff
@@ -0,0 +1,195 @@
...
+
+type decryptFunc func(gorp.SqlExecutor, int64, string) (string, error)
+
+// ParseAndImport parse an exportentities.Application and insert or update teh application in database
+func ParseAndImport(db gorp.SqlExecutor, cache cache.Store, proj *sdk.Project, eapp *exportentities.Application, force bool, decryptFunc decryptFunc, u *sdk.User) ([]sdk.Message, error) {
+	log.Info("ParseAndImport>> Import application %s in project %s (force=%v)", eapp.Name, proj.Key, force)
+	log.Debug("ParseAndImport>> App: %+v", eapp)
+
+	//Check valid application name
+	rx := regexp.MustCompile(sdk.NamePattern)
```
yesnault: regexp.MustCompile(sdk.NamePattern) is written 30x in CDS Code. Should be great if it's compile 1 time in sdk.

bnjjj: agree with that
+1

fsamin: ok will do in a next PR

PR #1640, opened 2017-12-05
feat: change workflow run number

sguiheux:


engine/api/workflow/dao_run.go
```diff
@@ -359,6 +359,40 @@ ORDER BY tags.tag;
...
+		if _, err := nextRunNumber(db, w); err != nil {
+			return sdk.WrapError(err, "UpdateRunNum> Cannot create run number")
+		}
+		return nil
+	}
+
+	query := `
+		UPDATE workflow_sequences set current_val = $1 WHERE workflow_id = $2
+	`
+	if _, err := db.Exec(query, num, w.ID); err != nil {
```
fsamin: we should check the current value to avoid set inferior num, with a select for update

engine/api/workflow_run.go
```diff
@@ -142,6 +142,61 @@ func (api *API) getWorkflowRunsHandler() Handler {
...
+		m := struct {
+			Num int64 `json:"num"`
+		}{
+			Num: num,
+		}
+		return WriteJSON(w, r, m, http.StatusOK)
+	}
+}
+
+func (api *API) postWorkflowRunNumHandler() Handler {
```
fsamin: please add a unit test

engine/api/workflow_run.go
```diff
@@ -142,6 +142,61 @@ func (api *API) getWorkflowRunsHandler() Handler {
 	}
 }
 
+func (api *API) getWorkflowRunNumHandler() Handler {
```
fsamin: please add or update a unit test like `TestGetWorkflowJobQueueHandler`

engine/api/workflow/dao_run.go
sguiheux: already done on handler

engine/api/workflow_run.go
sguiheux: done

engine/api/workflow_run.go
sguiheux: done

PR #1646, opened 2017-12-05
fix(api): delete lock on stop workflow run and stop worker when browse history

bnjjj:


ui/src/app/views/workflow/run/node/history/history.component.ts
```diff
@@ -18,6 +18,8 @@ export class WorkflowNodeRunHistoryComponent extends Table {
     @Input() currentBuild: WorkflowNodeRun;
     @Input() workflowName: string;
 
+    @Output() historyChanged = new EventEmitter<number>();
```
sguiheux: add but not used?

PR #1654, opened 2017-12-05
fix(cli): change env name to host instead api

bnjjj:


fsamin:
the commit history on this branch is really weird

i think CDS_HOST, only used in the old sdk, never has been clear enough: which host UI or API ? with or without http(s) ? with or without port ? how to to deal with a path ?

bnjjj:
In the cli documentation we mention several Times « host » we need to make a choice

bnjjj:
Update ! When merged please remind me to update cds it plan :)

PR #1677, opened 2017-12-10
fix (cdsctl): workflow pull & ask override files

yesnault:
os.Stat(fname) returns nil if file exists
add quiet flag, default false. User can see that there are many files written by cmd.

### DIFF
--- a/cli/cdsctl/workflow.go
+++ b/cli/cdsctl/workflow.go
@@ -383,7 +383,7 @@ func workflowPullRun(c cli.Values) error {
 			return err
 		}
 		if !c.GetBool("quiet") {
-			fmt.Printf("File %s created\n", fname)
+			fmt.Println(fname)
 		}
 	}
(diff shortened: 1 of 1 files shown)


cli/cdsctl/workflow.go
```diff
@@ -376,6 +382,9 @@ func workflowPullRun(c cli.Values) error {
 		if err := fi.Close(); err != nil {
 			return err
 		}
+		if !c.GetBool("quiet") {
+			fmt.Printf("File %s created\n", fname)
```
fsamin: why not just print the filename with the relative path from the working directory ? This way it can be easily used in a bash script

yesnault: yes

yesnault:
$cdsctl workflow pull 00SIM mywf
.cds/mywf.yml
.cds/mypip.pip.yml

PR #1678, opened 2017-12-10
feat (cdsctl): use workflow run count

yesnault:

### DIFF
--- a/engine/api/api_routes.go
+++ b/engine/api/api_routes.go
@@ -288,6 +288,7 @@ func (api *API) InitRouter() {
 
 	//Workflow queue
 	r.Handle("/queue/workflows", r.GET(api.getWorkflowJobQueueHandler))
+	r.Handle("/queue/workflows/count", r.GET(api.countWorkflowJobQueueHandler))
 	r.Handle("/queue/workflows/requirements/errors", r.POST(api.postWorkflowJobRequirementsErrorHandler, NeedWorker()))
 	r.Handle("/queue/workflows/{id}/take", r.POST(api.postTakeWorkflowJobHandler, NeedWorker()))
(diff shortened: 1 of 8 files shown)


yesnault:
We have to limit GET queue size too. This have to be done with a limit on each project to avoid that a project with 1k jobs does not let another job from another project running. cf. #1679

engine/api/workflow/dao_node_run_job.go
```diff
@@ -13,10 +13,11 @@ import (
 	"github.com/ovh/cds/engine/api/cache"
 	"github.com/ovh/cds/engine/api/group"
 	"github.com/ovh/cds/sdk"
+	"github.com/ovh/cds/sdk/log"
 )
 
-// LoadNodeJobRunQueue load all workflow_node_run_job accessible
-func LoadNodeJobRunQueue(db gorp.SqlExecutor, store cache.Store, groupsID []int64, since *time.Time, statuses ...string) ([]sdk.WorkflowNodeJobRun, error) {
+// return query, groupID, statuses, isSharedInfraGroup
+func queryLoadNodeJobRunQueue(isCount bool, groupsID []int64, since *time.Time, statuses ...string) (string, string, *time.Time, string, bool) {
```
fsamin: i don't like it, i would prefer two different functions event if there is a bit of code duplication 
- or may be groupID stuff should be in a function
- or the query should be in a constant

engine/api/api_routes.go
```diff
@@ -288,6 +288,7 @@ func (api *API) InitRouter() {
 
 	//Workflow queue
 	r.Handle("/queue/workflows", r.GET(api.getWorkflowJobQueueHandler))
+	r.Handle("/queue/workflows/count", r.GET(api.countWorkflowJobQueueHandler))
```
fsamin: I miss unit tests :)

yesnault: done

PR #1708, opened 2017-12-13
fix (api): remove verbose logs

fsamin:


engine/api/workflow/workflow_run_event.go
```diff
@@ -18,13 +18,11 @@ func GetWorkflowRunEventData(cError <-chan error, cEvent <-chan interface{}) ([]
 	for {
 		select {
 		case e, has := <-cError:
-			log.Info("GetWorkflowRunEventData> cError has: %t err:%v", has, e)
-			err = sdk.WrapError(e, "GetWorkflowRunEventData> Error received")
 			if !has {
+				err = sdk.WrapError(e, "GetWorkflowRunEventData> Error received")
```
sguiheux: we can loose err, keep it  above if

PR #1709, opened 2017-12-13
feat (vcs): send event to vcs

sguiheux:


engine/vcs/github/client_status.go
```diff
@@ -12,120 +12,188 @@ import (
...
+	if g.DisableStatus {
+		log.Warning("github.SetStatus>  ⚠ Github statuses are disabled")
 		return nil
 	}
 
-	if g.DisableStatus {
-		log.Warning("⚠ Github statuses are disabled")
+	var data statusData
+	var err error
+	switch event.EventType {
```
bnjjj: why not use .(type) ?

sguiheux: because i can't

PR #1719, opened 2017-12-14
fix (worker): upload artifact through temp url

fsamin:


sdk/artifact.go
```diff
@@ -324,36 +328,43 @@ func uploadArtifactWithTempURL(project, pipeline, application, env, tag string,
...
+	for i := 0; i < retry; i++ {
+		path = fmt.Sprintf("/project/%s/application/%s/pipeline/%s/%d/artifact/%s/url/callback?envName=%s", project, application, pipeline, buildNumber, tag, url.QueryEscape(env))
+		_, _, globalErr = Request("POST", path, b)
+		if globalErr == nil {
+			return true, time.Since(t0), nil
+		}
 	}
 
-	return nil
+	return true, time.Since(t0), err
```
bnjjj: isn't about globalErr instead of err ?

PR #1732, opened 2017-12-15
fix(api): return error if we want to start a node run with checkCondi…

### LINKED ISSUE #1653: fix(api): return error if we want to start a node run with checkCondition to false

bnjjj:
…tion to false (close #1653)


ui/src/assets/i18n/en.json
```diff
@@ -679,6 +679,6 @@
   "workflow_orientation_vertical" : "Vertical",
   "workflow_orientation_tree" : "Tree",
   "workflow_last_execution" : "Last execution date",
-  "workflow_first_execution" : "First execution date"
-
+  "workflow_first_execution" : "First execution date",
+  "workflow_runnumber_title": "First run number of workflow"
```
sguiheux: Current run number

PR #1735, opened 2017-12-16
feat (api): default payload

yesnault:
when we add a RepositoryWebhook and default payload empty

### DIFF
--- a/sdk/workflow.go
+++ b/sdk/workflow.go
@@ -640,6 +640,11 @@ type WorkflowNodeContext struct {
 	Conditions                WorkflowNodeConditions `json:"conditions,omitempty" db:"-"`
 }
 
+type WorkflowNodeContextDefaultPayloadVCS struct {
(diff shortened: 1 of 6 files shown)


sdk/workflow.go
```diff
@@ -640,6 +640,11 @@ type WorkflowNodeContext struct {
 	Conditions                WorkflowNodeConditions `json:"conditions,omitempty" db:"-"`
 }
 
+type WorkflowNodeContextDefaultPayloadVCS struct {
```
sguiheux: doc

sdk/workflow.go
```diff
@@ -640,6 +640,11 @@ type WorkflowNodeContext struct {
...
+	GitBranch string `json:"git.branch" db:"-"`
+	GitHash   string `json:"git.hash" db:"-"`
+}
```
sguiheux: missing git.author, git.hash.before, git.repository, giut.message

PR #1737, opened 2017-12-17
fix (engine): build freebsd

yesnault:
it's a temporary patch, should be reviewed when we go on deps (and so, removing docker vendor, for using moby)


yesnault:
(see gitter discussion)

yesnault:
it does not affect linux and windows

PR #1738, opened 2017-12-17
feat(ui): add priority order to display tags in order in sidebar

bnjjj:


bnjjj:
close #1674

PR #1789, opened 2017-12-23
fix(ui): add scroll in list commit in modal run #1784

bnjjj:


bnjjj:
close #1784

PR #1796, opened 2017-12-24
fix (hooks): update / delete scheduler

yesnault:
When a user delete or udpate a scheduler, we have to restart service hooks to take care of modification.

The func runTasks(...) was not call as a long-running goroutine, it's now sync each 30s.

### DIFF
--- a/engine/api/workflow/hook.go
+++ b/engine/api/workflow/hook.go
@@ -107,7 +107,6 @@ func deleteHookConfiguration(db gorp.SqlExecutor, store cache.Store, p *sdk.Proj
 				}
 			}
 		}
-		return nil
 	}
 
 	//Push the hook to hooks µService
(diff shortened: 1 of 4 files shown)


yesnault:
@fsamin feel free to update this PR directly if needed.
cc @admdwrf it will fix your issue about updating the scheduler on a CDS Worfklow

fsamin:
Indeed restart the service or simulating the restart with this PR solves the problem because it performs hook resync. But we should not fix it like this

But i think the bug is on API side; somewhere in https://github.com/ovh/cds/blob/master/engine/api/workflow/hook.go#L17. I guess the hooks resync fails with schedulers

yesnault:
@fsamin thks, I'll fix

yesnault:
@fsamin should we put s.runTasks(ctx) on a for loop (each 10min?)

fsamin:
Well, i don't think so. 
On hooks startup, all tasks are synchronized with hooks and started. but then tasks have to be managed (stop, start, update, ; delete) only from the Hooks REST API; 

The role of the hook registration on workflow is to synchronize all the hooks of a workflow when it is updated.

yesnault:
ok, thks

engine/api/workflow/hook.go
```diff
@@ -119,7 +117,7 @@ func deleteHookConfiguration(db gorp.SqlExecutor, store cache.Store, p *sdk.Proj
 	}
 	code, errHooks := services.DoJSONRequest(srvs, http.MethodDelete, fmt.Sprintf("/task/bulk"), hookToDelete, nil)
 	if errHooks != nil || code >= 400 {
-		log.Warning("HookRegistration> Unable to delete old hooks [%d]: %s", code, errHooks)
+		return fmt.Errorf("HookRegistration> Unable to delete old hooks [%d]: %s", code, errHooks)
```
sguiheux: if you return an error, transaction will be rollbacked => hook will in database be not anymore on gitlab/bitbucket/github.

yesnault:
@fsamin ready to merge I think

PR #1805, opened 2017-12-27
feat(api): add vcs infos into a node run and refactor get build commits

bnjjj:


engine/sql/072_workflow_node_run_vcs.sql
```diff
@@ -0,0 +1,8 @@
+-- +migrate Up
```
sguiheux: Why not a data column JSONB and use specific json request ?
A nodeRun must not have specific column about repository (that's  why we create workflow :))

bnjjj: Discussed with @fsamin about jsonb infos. The fact is that query some data in jsonb could be painful and slow. And considering informations about vcs, if no application is linked we don’t use these data and it stays empty. The conclusion of this solution is that relying on tag to fetch commit and other informations is a bad thing. Tag should stay to inform users and we have not to use these for our process in our code

sguiheux: not sure about paintful and slow if we consider a simple map object.
I agree about using tags.

engine/api/workflow/dao_node.go
```diff
@@ -266,6 +266,26 @@ func loadNode(db gorp.SqlExecutor, store cache.Store, w *sdk.Workflow, id int64,
 	return &wn, nil
 }
 
+// LoadNodeContextByNodeName load the context for a given node name and user
+func LoadNodeContextByNodeName(db gorp.SqlExecutor, store cache.Store, nodeName string) (*sdk.WorkflowNodeContext, error) {
+	dbnc := NodeContext{}
+	query := `
+		SELECT workflow_node_context.id, workflow_node_context.workflow_node_id FROM workflow_node_context
+		JOIN workflow_node ON workflow_node.id = workflow_node_context.workflow_node_id
+		WHERE workflow_node.name = $1
```
sguiheux: missing condition on workflow_id.  Name is not unique in workflow_node

engine/api/workflow/dao_node_run.go
```diff
@@ -418,33 +429,42 @@ func GetNodeRunBuildCommits(db gorp.SqlExecutor, store cache.Store, p *sdk.Proje
...
-		JOIN workflow_node_run ON workflow_node_run.workflow_run_id = wrt1.workflow_run_id
-		JOIN workflow_node_context ON workflow_node_context.workflow_node_id = workflow_node_run.workflow_node_id
-		WHERE wrt1.tag = 'git.branch'
-		AND workflow_node_run.status = 'Success'
-		AND workflow_node_run.workflow_node_id = $1
-		AND workflow_node_run.num < $2
-		AND workflow_node_context.application_id = $3
+		SELECT workflow_node_run.vcs_branch, workflow_node_run.vcs_hash, workflow_node_run.vcs_repository, workflow_node_run.num
+		FROM workflow_node_run
+		JOIN workflow_node ON workflow_node.name = $1
```
sguiheux: not enough for JOIN, Name is not unique

PR #1806, opened 2017-12-27
feat: display os / archi with /version

yesnault:

### DIFF
--- a/cli/cdsctl/Makefile
+++ b/cli/cdsctl/Makefile
@@ -4,7 +4,7 @@ VERSION := $(if ${CDS_SEMVER},${CDS_SEMVER},snapshot)
 
 TARGET_DIR = ./dist
 TARGET_BINARY = cdsctl
-TARGET_LDFLAGS = -ldflags "-X github.com/ovh/cds/sdk.VERSION=$(VERSION)"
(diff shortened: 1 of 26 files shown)


fsamin:
Why injecting them and not using `runtime.GOOS` and `runtime.GOARCH` ?

https://golang.org/pkg/runtime/#pkg-constants

fsamin:
And what is the goal of OS and ARCH ?

yesnault:
@fsamin  +1 for runtime.GOOS and runtime.GOARCH (pr updated) with new fields in graylog too.

Goal: best debug context when we'll have multi-arch Hatcheries / Plugin / Workers. When we ask user to run /version, we'll have all useful data about the version.

yesnault:
```
./plugin-download version
CDS Plugin version:snapshot os:darwin architecture:amd64
./engine version
CDS API version:snapshot os:darwin architecture:amd64
./worker version
CDS Worker version:snapshot os:darwin architecture:amd64
```

fsamin:
@yesnault okay, and it would be more interesting to exploit data sent from worker (and hatcheries should do it as well) https://github.com/ovh/cds/blob/master/engine/worker/register.go#L13

yesnault:
@fsamin yes, some refacto in pr updated + add version/os/arch in spawn infos (wf only). Multi-arch will be done in others PRs (plugins, then workers, then hatcheries)

PR #1848, opened 2018-01-03
feat (plugin): use sdk.Interpolate

yesnault:


contrib/plugins/plugin-marathon/marathon.go
```diff
@@ -289,7 +290,7 @@ func tmplApplicationConfigFile(a plugin.IJob, filepath string) (string, error) {
 	}
 
 	// apply cds.var on marathon.json file
-	out, errapp := plugin.ApplyArguments(a.Arguments().Data, buff)
+	out, errapp := sdk.Interpolate(string(buff), a.Arguments().Data)
```
fsamin: humm, are we sure we want to load the sdk package (which try to init a connection to the API) in the plugin ?

PR #1851, opened 2018-01-04
feat (cdsctl): refactor config storage and keychain

fsamin:


cli/cdsctl/configstore_linux_amd64.go
```diff
@@ -0,0 +1,34 @@
+package main
+
+import (
+	"io"
+
+	"github.com/naoina/toml"
```
bnjjj: space after

fsamin: updated

PR #1855, opened 2018-01-04
feat(api,ui): add the ability to run multiple node on the same level …

### LINKED ISSUE #1790: feat(api): add "can_run" info for workflow node in a workflow run

bnjjj:
… close #1790

### DIFF
--- a/engine/api/workflow/dao_run.go
+++ b/engine/api/workflow/dao_run.go
@@ -314,6 +314,7 @@ func loadRun(db gorp.SqlExecutor, withArtifacts bool, query string, args ...inte
 	return &wr, nil
 }
 
+//TODO: if no bugs are found, it could be used to refactor process.go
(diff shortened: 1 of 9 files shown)


engine/api/workflow/dao_run.go
```diff
@@ -313,6 +314,40 @@ func loadRun(db gorp.SqlExecutor, withArtifacts bool, query string, args ...inte
 	return &wr, nil
 }
 
+// canBeRun return boolean to know if a wokrflow node run can be run or not
```
fsamin: should we refactor the workflow.Process function to call this ?

bnjjj: Ok, I suggest testing for a moment in production (it's less critical) and if it's stable it could be put in process.go to make some refactor :) 
I've added a TODO

fsamin: I'm fine with it

PR #1869, opened 2018-01-05
fix (ui):  add load workflow loader

sguiheux:


ui/src/app/views/workflow/show/workflow.html
```diff
@@ -117,5 +117,5 @@ <h3>{{ 'workflow_permission_form_title' | translate }}</h3>
     </div>
 </ng-template>
 <ng-template #loadingPart>
-    <div class="ui text active loader" *ngIf="!detailedWorkflow || !project">Loading workflow...</div>
+    <div class="ui text active loader">Loading workflow...</div>
```
bnjjj: {{ 'common_loading_workflow' | translate }}

PR #1882, opened 2018-01-05
fix (sdk): espace tag on Container Swift

yesnault:
This will avoid this:

SwiftStore> Creating new session key for 1111-11111-%!B(MISSING)%!B(MISSING).git.branch%!D(MISSING)%!D(MISSING)trusty5
and error from switf after


fsamin:
Why the {{.git.branch}} variable wasn't replaced ? I suggest to forbid uninterpolated `{{.*}}` as parameter value in builtin actions. This was also the root cause of #1510 :/

With this fix, i think it could hide some weird behavior with artifacts

PR #1884, opened 2018-01-06
feat (api): workflow Mutex

fsamin:
Adding a mutex on a workflow node will let the node run at status `waiting` if the node is already running somewhere else.

Major impacts:

* Persist mutex on workflow node in database
* Added workflow node name in the workflow run for conveniency
* If there is a mutex on a node, avoid call `workflow.execute` on the node run
* At the end of workflow node run, if there is a mutex, try to find other workflow node run for the same node and trigger `workflow.execute` on it

Next:

* We have to discuss about how to manage event channel on mutex release
* UI features will come in another PR

### DIFF
--- a/engine/sql/073_workflow_mutex.sql
+++ b/engine/sql/073_workflow_mutex.sql
@@ -1,0 +1,11 @@
+-- +migrate Up
+ALTER TABLE workflow_node_context ADD COLUMN mutex BOOLEAN DEFAULT FALSE;
+UPDATE workflow_node_context SET mutex = 'false';
+
+ALTER TABLE workflow_node_run ADD COLUMN workflow_node_name text;
+UPDATE workflow_node_run SET workflow_node_name = '';
+
+-- +migrate Down
+ALTER TABLE workflow_node_context DROP COLUMN mutex;
+ALTER TABLE workflow_node_run DROP COLUMN workflow_node_run;
+
--- a/engine/sql/073_workflow_mutex.sql
+++ b/engine/sql/073_workflow_mutex.sql
@@ -2,7 +2,7 @@
 ALTER TABLE workflow_node_context ADD COLUMN mutex BOOLEAN DEFAULT FALSE;
 UPDATE workflow_node_context SET mutex = 'false';
 
-ALTER TABLE workflow_node_run ADD COLUMN workflow_node_name text;
+ALTER TABLE workflow_node_run ADD COLUMN workflow_node_name text DEFAULT '';
 UPDATE workflow_node_run SET workflow_node_name = '';
 
 -- +migrate Down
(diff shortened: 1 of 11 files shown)


engine/sql/073_workflow_mutex.sql
```diff
@@ -0,0 +1,11 @@
+-- +migrate Up
+ALTER TABLE workflow_node_context ADD COLUMN mutex BOOLEAN DEFAULT FALSE;
+UPDATE workflow_node_context SET mutex = 'false';
+
+ALTER TABLE workflow_node_run ADD COLUMN workflow_node_name text;
```
yesnault: DEFAULT ''

sdk/workflow_run.go
```diff
@@ -92,6 +92,7 @@ type WorkflowNodeRun struct {
 	WorkflowRunID      int64                            `json:"workflow_run_id"`
 	ID                 int64                            `json:"id"`
 	WorkflowNodeID     int64                            `json:"workflow_node_id"`
+	WorkflowNodeName   string                           `json:"workflow_node_string"`
```
yesnault: workflow_node_string -> workflow_node_name, no ?

sdk/messages.go
```diff
@@ -69,6 +69,8 @@ var (
 	MsgWorkflowStarting                    = &Message{"MsgWorkflowStarting", trad{FR: "Le workflow %s#%s a été démarré", EN: "Workflow %s#%s has been started"}, nil}
 	MsgWorkflowError                       = &Message{"MsgWorkflowError", trad{FR: "Une erreur est survenue: %v", EN: "An error has occured: %v"}, nil}
 	MsgWorkflowNodeStop                    = &Message{"MsgWorkflowNodeStop", trad{FR: "Le pipeline a été arrété par %s", EN: "The pipeline has been stopped by %s"}, nil}
+	MsgWorkflowNodeMutex                   = &Message{"MsgWorkflowNodeMutex", trad{FR: "Le pipeline %s est en attente de la libération du mutex", EN: "The pipeline %s is waiting for mutex release"}, nil}
```
sguiheux: Don't think that mutex is a good word for user experience. Maybe we have to say that this node is currently runnning on  another run

sguiheux:
Missing UI impact

sguiheux:
By adding workflow node name on workflow node run, all codes that try to get previous node may have to be refact removing join with workflow_node (Notification, Commits )=> have to create Issue

sdk/messages.go
fsamin: I don't how to represent/call it on UI also ? Any idea ?

fsamin:
@sguiheux 

As I said, UI impact will come in another PR. Without UI impacts; the behavior in unchanged.
Impact on Workflow export is also missing; but it's not a problem since the behavior in unchanged.

=> #1887

sdk/messages.go
sguiheux: "Ce pipeline ne peut être lancé tant qu'il est en cours sur un autre run. Mise en attente.."

Release:  "Lancement du pipeline"

No specific representation on UI, waiting status (with the clock icon) is enough i think.

Issue #870, opened 2017-07-20
Hook on worker update


henyxia:
Hi,

I'd like to have the ability to register something (like an HTTP hook) to be notified when your worker is updated.

Thanks

PR #1885, opened 2018-01-06
feat (worker): worker auto-update

yesnault:
New flag on the main cmd: --auto-update

the command ./worker [args] -> internally os.exec worker run [args]. If auto-update -> update the worker before launch the run cmd. If auto-update -> the run command check each 10min the version on api. If version is != "" : sdk.Exit and the main process will relaunch the worker.

No impact on user / hatchery.

close https://github.com/ovh/cds/issues/870

### DIFF
--- a/engine/worker/cmd_main.go
+++ b/engine/worker/cmd_main.go
@@ -103,7 +103,6 @@ func mainCommandRun(w *currentWorker) func(cmd *cobra.Command, args []string) {
 
 		if cmd.Flag("auto-update").Value.String() == "true" {
 			updateCmd(w)(cmd, args)
-			log.Info("Restart worker done") // fake msg
 		}
 		toRun := true
 		for toRun {
--- a/engine/worker/cmd_main.go
+++ b/engine/worker/cmd_main.go
@@ -1,8 +1,6 @@
 package main
 
 import (
-	"bufio"
-	"fmt"
 	"os"
 	"os/exec"
 
@@ -126,55 +124,14 @@ func execWorker() {
 	args := []string{"run"}
 	args = append(args, os.Args[1:]...)
 	cmd := exec.Command(current, args...)
-	stdout, err := cmd.StdoutPipe()
-	if err != nil {
-		sdk.Exit("Error on starting worker (StdoutPipe)", err)
-	}
-	stderr, err := cmd.StderrPipe()
-	if err != nil {
-		sdk.Exit("Error on starting worker (StderrPipe)", err)
-	}
 
-	stdoutreader := bufio.NewReader(stdout)
-	stderrreader := bufio.NewReader(stderr)
-
-	outchan := make(chan bool)
-	go func() {
-		for {
-			line, errs := stdoutreader.ReadString('\n')
-			if errs != nil {
-				stdout.Close()
-				close(outchan)
-				return
-			}
-			if line != "" {
-				fmt.Printf(line)
-			}
-		}
-	}()
-
-	errchan := make(chan bool)
-	go func() {
-		for {
-			line, errs := stderrreader.ReadString('\n')
-			if errs != nil {
-				stderr.Close()
-				close(errchan)
-				return
-			}
-			if line != "" {
-				fmt.Printf(line)
-			}
-
-		}
-	}()
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
 
 	if err := cmd.Start(); err != nil {
 		log.Error("start err:%s", err)
 	}
 
-	<-outchan
-	<-errchan
 	if err := cmd.Wait(); err != nil {
 		log.Error("wait err:%s", err)
 	}
--- a/engine/worker/cmd_main.go
+++ b/engine/worker/cmd_main.go
@@ -1,15 +1,11 @@
 package main
 
 import (
-	"context"
-	"encoding/json"
+	"bufio"
 	"fmt"
 	"os"
-	"os/signal"
-	"syscall"
-	"time"
+	"os/exec"
 
-	"github.com/google/gops/agent"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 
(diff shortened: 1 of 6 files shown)


engine/worker/cmd_run.go
```diff
@@ -0,0 +1,360 @@
...
+
+func cmdRun(w *currentWorker) *cobra.Command {
+	c := &cobra.Command{
+		Use:   "run",
+		Short: "worker run.",
+		Long:  "worker run: you should not need this command directly",
+		Run:   runCmd(w),
+	}
+
+	initFlagsRun(c)
```
fsamin: why duplicating the flags and not used persistent flags of the root command ?

engine/worker/cmd_register.go
```diff
@@ -30,7 +30,7 @@ func cmdRegisterRun(w *currentWorker) func(cmd *cobra.Command, args []string) {
 			ModelID:      w.model.ID,
 		}
 
-		if err := w.register(form); err != nil {
+		if err := w.register(form, viper.GetBool("auto_update")); err != nil {
```
fsamin: why do we want to run `worker register --auto-update` ?
i suppose we can do register with auto-update to false here

engine/worker/cmd_run.go
```diff
@@ -0,0 +1,360 @@
...
+			if err := agent.Listen(nil); err != nil {
+				sdk.Exit("Error on starting gops agent", err)
+			}
+		}
+
+		hostname, errh := os.Hostname() // no check of err here
+		if errh != nil {
+			hostname = fmt.Sprintf("error compute hostname: %s", errh)
+		}
+		log.Info("What a good time to be alive, I'm in version %s, my hostname is %s", sdk.VERSION, hostname)
```
fsamin: Why not have a cleaner log ?

````
CDS Worker starting
version: %s
hostname: %s
auto-update: %v
````

File: engine/worker/cmd_run.go [hunk 4]
```diff
@@ -0,0 +1,360 @@
...
+
+		if viper.GetString("log_level") == "debug" {
+			if err := agent.Listen(nil); err != nil {
+				sdk.Exit("Error on starting gops agent", err)
+			}
+		}
+
+		hostname, errh := os.Hostname() // no check of err here
+		if errh != nil {
+			hostname = fmt.Sprintf("error compute hostname: %s", errh)
```
fsamin: why not exit ? it's a weird hostname

engine/worker/cmd_main.go
```diff
@@ -90,318 +94,89 @@ func cmdMain(w *currentWorker) *cobra.Command {
...
-				os.Exit(1)
-			}
-			if err := w.doRegister(); err != nil {
-				log.Error("Unable to register to CDS (%v). Retry", err)
+		viper.SetEnvPrefix("cds")
+		viper.AutomaticEnv()
+
+		if cmd.Flag("auto-update").Value.String() == "true" {
+			updateCmd(w)(cmd, args)
+			log.Info("Restart worker done") // fake msg
```
fsamin: lol, what a mess !

engine/worker/cmd_main.go
```diff
@@ -90,318 +94,89 @@ func cmdMain(w *currentWorker) *cobra.Command {
...
 
-		// start queue polling
-		pbjobs := make(chan sdk.PipelineBuildJob, 1)
-		wjobs := make(chan sdk.WorkflowNodeJobRun, 1)
-		errs := make(chan error, 1)
+	log.Info("Current binary: %s", current)
+	args := []string{"run"}
+	args = append(args, os.Args[1:]...)
+	cmd := exec.Command(current, args...)
+	stdout, err := cmd.StdoutPipe()
```
fsamin: why not set the `cmd.StdOut` to `os.StdOut` to avoid buffering outputs

engine/worker/cmd_main.go
```diff
@@ -90,318 +94,89 @@ func cmdMain(w *currentWorker) *cobra.Command {
...
+	if err != nil {
+		sdk.Exit("Error on starting worker (StdoutPipe)", err)
+	}
+	stderr, err := cmd.StderrPipe()
```
fsamin: idem for `os.SdtErr`

engine/worker/cmd_main.go
```diff
@@ -90,318 +94,89 @@ func cmdMain(w *currentWorker) *cobra.Command {
...
-			log.Info("Enter endFunc")
-			w.drainLogsAndCloseLogger(ctx)
-			registerTick.Stop()
-			w.unregister()
-			cancel()
-
-			if viper.GetBool("force_exit") {
-				log.Info("Exiting worker with force_exit true")
+	outchan := make(chan bool)
+	go func() {
```
fsamin: this should be useless with above suggestions

engine/worker/cmd_run.go
yesnault: The persistent flags are persistent again. But flags

```
single-use
auto-update
force-exit
basedir
ttl
booked-pb-job-id
booked-workflow-job-id
booked-job-id
grpc-api
grpc-insecure
graylog-protocol
graylog-host
graylog-port
graylog-extra-key
graylog-extra-value
```
are not useful to other commands than the root/run commands.

Duplicate theses flags between root cmd and run cmd is usefull to run sub-process with

```golang
current, errExec := os.Executable()
args := []string{"run"}
args = append(args, os.Args[1:]...)
cmd := exec.Command(current, args...)
```

fsamin: ok, got it

fsamin:
Seems nice, but about worker cancellation (eg Ctrl-C), the worker catches the signal to cancel gracefully the tasks by cancelling the root context. Should the root process catch the signal and send it to the child process or it's done by OS ?

yesnault:
Ctrl+C kill the root process, but the signal is catched before by the sub process for a graceful kill.

PR #1886, opened 2018-01-06
refactor (*): remove templates 🚧

fsamin:
don't merge


sguiheux:
Don't forget UI Impact => create an empty application call Void template

fsamin:
indeed

ui/package-lock.json
```diff
@@ -1635,6 +1635,7 @@
       "requires": {
         "anymatch": "1.3.2",
         "async-each": "1.0.1",
+        "fsevents": "1.1.3",
```
yesnault: there a link with templates?

fsamin: i don't think so but every npm install i ran this last week end added this

PR #1892, opened 2018-01-08
fix(api): fix update build commits

bnjjj:


engine/api/workflow/resync_workflow.go
```diff
@@ -88,3 +89,60 @@ func ResyncWorkflowRunStatus(db gorp.SqlExecutor, wr *sdk.WorkflowRun, chEvent c
...
+			}
+
+			commits, curVCSInfos, err := GetNodeRunBuildCommits(db, store, proj, &wr.Workflow, n.Name, wr.Number, &nr, n.Context.Application, n.Context.Environment)
+			if err != nil {
+				log.Error("ResyncNodeRuns> cannot get build commits on a node run %v", err)
+			} else {
+				nr.Commits = commits
+			}
+
+			if commits != nil {
```
fsamin: I think here, it cannot be nil

engine/api/workflow/resync_workflow.go
```diff
@@ -88,3 +89,60 @@ func ResyncWorkflowRunStatus(db gorp.SqlExecutor, wr *sdk.WorkflowRun, chEvent c
 
 	return nil
 }
+
+// ResyncNodeRunsWithCommits load commits build in this node run and save it into node run
+func ResyncNodeRunsWithCommits(db gorp.SqlExecutor, store cache.Store, proj *sdk.Project, nodeRuns []sdk.WorkflowNodeRun) {
+	for _, nodeRun := range nodeRuns {
+		if nodeRun.Commits != nil {
```
fsamin: do you want really want t check slice is nil or len(slice) is 0 ?

PR #1906, opened 2018-01-08
tests (*): simple workflow scenario

fsamin:


fsamin:
workflow will succeed after https://github.com/ovh/cds/pull/1904

tests/sc_workflow_create_simple.yml
```diff
@@ -0,0 +1,28 @@
+name: Create a simple workflow (ITSCWRKFLW1) and pull it
```
yesnault: ITSCWRKFLW1 -> ITSCWRKFLW1-{{.cds.version}}.{{.cds.run.subnumber}} to allow running IT test on many branches at the same time (or running it even if previous run was failed on this run)

fsamin: The goal of this test is not to run the workflow.
It's only testing workflow creation and pull from the CLI.

the test starts by removing all the projects; it can be run again and again and again without troubles

it will push a few other test cases

fsamin: and i want to keep the test very simple to be able to run it locally just with the `cds.build.cdsctl` variable

yesnault: Ok. We will refactor IT Tests globally to make all tests running on a uniq prj key / group at the same time.  If a
+  - script: "{{.cds.build.cdsctl}} project remove --force ITSCWRKFLW1"
is done (on an other run) after a
+  - script: "{{.cds.build.cdsctl}} project add ITSCWRKFLW1 ITSCWRKFLW1"
it will fail (with a Mutex, it should be ok).

fsamin: Not sure it's a good idea; this kind of tests deserves stability and deterministic behavior.
Randomness of parallelism and changing data can change tests results; which is not good at all.

I also prefer to use static inputs to be able to check result manually if need to => it means we shouldn't remove test data and results after at the end of the testsuites.

I think we doesn't need to add complexity in the tests to be able to run them in parallels. Th mutex if enough to avoid false failing tests.
If we run integration tests on a single environment, It sounds natural to set the mutex on.

We will have better benefits if we spend more time on business scope of the tests, rather on hacking them

PR #1929, opened 2018-01-11
fix(api): set node run skipped if all stages are skipped

bnjjj:


bnjjj:
don't merge now ! I detect another bug :)

bnjjj:
ready to merge with new fix (when your last stage of your build is disabled or skipped)

PR #1930, opened 2018-01-11
feat (api,worker,ui): new os-architecture requirement

### LINKED ISSUE #549: feat (api): os requirement

fsamin:
Impacts on worker model, hatcheries, plugins will come later in other PR

Close #549, #550


ui/src/app/shared/requirements/form/requirements.form.component.ts
```diff
@@ -120,6 +128,9 @@ export class RequirementsFormComponent implements OnInit {
             case 'volume':
                 this.newRequirement.name = this.getVolumeName();
                 break;
+            case 'os-architecture':
```
bnjjj: could be interesting to create a const variable in this class in order to not mispell the requirement type. It's only used in this class I think and in the template

ui/src/app/shared/requirements/list/requirements.list.component.ts
```diff
@@ -34,23 +34,32 @@ export class RequirementsListComponent extends Table {
...
+    _suggestWithOSArch: Array<string> = [];
+
     loading = true;
 
     constructor(private _requirementStore: RequirementStore, private _workerModelService: WorkerModelService) {
         super();
         this.nbElementsByPage = 5;
+    }
+
+    ngOnInit() {
```
bnjjj: I think we don't need to put in ngOnInit(), because you don't use values from input inside

sguiheux:
Do you have a linter problem?

ui/src/app/shared/requirements/list/requirements.list.component.ts
fsamin: i don't know, i just found it easier to have to same behavior, code in the list component and in the form component

fsamin:
No i think, you have an IDE problem :)

PR #1944, opened 2018-01-13
fix (worker): artifact upload - check errors

yesnault:
without waiting the end of append in globalError,
globalError could be empty after <-chanError: the test
of empty could be done before than the goroutine with sendLog + Append.

Add a waitGroup ensure that error in well append in globalError before checking
that globalError is empty or not.


yesnault:
Playground: https://play.golang.org/p/NNfcK6SzslG

PR #1945, opened 2018-01-13
feat (api,ui): improve download page.

yesnault:
Display if binary is available or not on UI


yesnault:
Example "If there is only one binary available" --> 

capture d ecran 2018-01-13 a 17 38 48

yesnault:
old routes: I have to check with users before remove them (it's on board)

PR #1989, opened 2018-01-18
feat (api): get all workflow runs

fsamin:


engine/api/api_routes.go
```diff
@@ -208,6 +208,7 @@ func (api *API) InitRouter() {
 	r.Handle("/project/{permProjectKey}/push/workflows", r.POST(api.postWorkflowPushHandler))
 
 	// Workflows run
+	r.Handle("/project/{permProjectKey}/runs/", r.GET(api.getWorkflowAllRunsHandler))
```
sguiheux: remove last /

PR #1991, opened 2018-01-18
fix (api): duplicate git variables

fsamin:


engine/api/workflow/process.go
```diff
@@ -492,17 +492,58 @@ func processWorkflowNodeRun(dbCopy *gorp.DbMap, db gorp.SqlExecutor, store cache
 	}
 
 	//If git values doesn't exist in jobparams, they doesn't exist in gitValues; inject them from VCSInfos in run.BuildParameters
-	if v := gitValues[tagGitRepository]; v == "" {
-		sdk.AddParameter(&run.BuildParameters, tagGitRepository, sdk.StringParameter, run.VCSRepository)
+	if v, has := gitValues[tagGitRepository]; v == "" {
```
bnjjj: as the process.go and `processWorkflowNodeRun` is already long I think it could be great to have a function and don't duplicate same lines of code

PR #2020, opened 2018-01-20
chore: dep

yesnault:


fsamin:
👍 
btw i think we should remove iffy, it was a bad idea we already have too much test dependencies

yesnault:
@fsamin yes, totally agree. I'll test venom lib, it's more readable (proposition incoming in another PR). Removing iffy implies a tons of impacts, it will be probably in another PR for keeping this PR focusing on dep. This PR is already unreadable now :-)

yesnault:
@fsamin for iffy, wip on https://github.com/ovh/cds/tree/ye-iffy

fsamin:
yep for sure, no hurry to remove iffy :)
using venom lib should be great, but we have to check; i not sure i implemented the httprecorder stuff

PR #2065, opened 2018-01-25
feat(hatchery): add kubernetes hatchery

bnjjj:
Need more test with a real kubernetes cluster (not in local)


engine/hatchery/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,307 @@
...
+	"os"
+	"strings"
+	"time"
+
+	"github.com/ovh/cds/sdk"
+	"github.com/ovh/cds/sdk/cdsclient"
+	"github.com/ovh/cds/sdk/hatchery"
+	"github.com/ovh/cds/sdk/log"
+
+	"github.com/moby/moby/pkg/namesgenerator"
```
fsamin: don't we have our own namesgenerator ?

engine/hatchery/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,307 @@
...
+	if hconfig.API.HTTP.URL == "" {
+		return fmt.Errorf("API HTTP(s) URL is mandatory")
+	}
+
+	if hconfig.API.Token == "" {
+		return fmt.Errorf("API Token URL is mandatory")
+	}
+
+	if hconfig.Name == "" {
+		return fmt.Errorf("please enter a name in your local hatchery configuration")
```
fsamin: k8s

engine/hatchery/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,307 @@
...
+	return h.client
+}
+
+//Configuration returns Hatchery CommonConfiguration
+func (h *HatcheryKubernetes) Configuration() hatchery.CommonConfiguration {
+	return h.Config.CommonConfiguration
+}
+
+// ModelType returns type of hatchery
+func (*HatcheryKubernetes) ModelType() string {
```
fsamin: this should be the type of worker model ?

engine/hatchery/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,307 @@
...
+	return sdk.Kubernetes
+}
+
+// CanSpawn return wether or not hatchery can spawn model.
+// requirements are not supported
+func (h *HatcheryKubernetes) CanSpawn(model *sdk.Model, jobID int64, requirements []sdk.Requirement) bool {
+	for _, r := range requirements {
+		if r.Type == sdk.ServiceRequirement {
+			return false
```
fsamin: i'm so sad

engine/hatchery/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,307 @@
...
+			RestartPolicy:                 apiv1.RestartPolicyNever,
+			TerminationGracePeriodSeconds: &gracePeriodSecs,
+			Containers: []apiv1.Container{
+				{
+					Name:  name,
+					Image: spawnArgs.Model.Image,
+					Env:   envs,
+					Resources: apiv1.ResourceRequirements{
+						Requests: apiv1.ResourceList{
+							apiv1.ResourceMemory: resource.MustParse(fmt.Sprintf("%d", h.Config.DefaultMemory)),
```
fsamin: we could handle the memory requirement ?

engine/hatchery/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,307 @@
...
+			return false
+		}
+	}
+	return true
+}
+
+// SpawnWorker starts a new worker process
+func (h *HatcheryKubernetes) SpawnWorker(spawnArgs hatchery.SpawnArguments) (string, error) {
+	name := fmt.Sprintf("k8s-%s-%s", strings.ToLower(spawnArgs.Model.Name), strings.Replace(namesgenerator.GetRandomName(0), "_", "-", -1))
+	label := "execution"
```
fsamin: why "execution" ?

engine/hatchery/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,307 @@
...
+		},
+	})
+
+	return pod.Name, err
+}
+
+// WorkersStarted returns the number of instances started but
+// not necessarily register on CDS yet
+func (h *HatcheryKubernetes) WorkersStarted() int {
+	list, err := h.k8sClient.CoreV1().Pods(h.Config.KubernetesNamespace).List(metav1.ListOptions{LabelSelector: LABEL_WORKER})
```
fsamin: should we list by the hatchery name (and set the hatchery name as a label) ?

engine/hatchery/kubernetes/kubernetes.go
yesnault: hum, not since migrate to dep :-/

engine/hatchery/kubernetes/kubernetes.go
bnjjj: First iteration

engine/hatchery/kubernetes/kubernetes.go
bnjjj: Not really inspired :p any suggestion ?

engine/hatchery/kubernetes/kubernetes.go
bnjjj: Yes i forgot

bnjjj:
No @fsamin deps aren’t already there. I don’t want to put fire to my train’s wifi 😬

engine/hatchery/kubernetes/kubernetes.go
bnjjj: Yes I can add another label for hatchery name

engine/hatchery/kubernetes/kubernetes.go
bnjjj: :(

fsamin: don't we have our own namesgenerator package ?

bnjjj:
F**** DEP !!!

engine/hatchery/kubernetes/kubernetes.go
bnjjj: done

engine/hatchery/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,331 @@
...
+		}
+	}
+
+	return workersLen
+}
+
+// WorkersStartedByModel returns the number of instances of given model started but
+// not necessarily register on CDS yet
+func (h *HatcheryKubernetes) WorkersStartedByModel(model *sdk.Model) int {
+	list, err := h.k8sClient.CoreV1().Pods(h.Config.KubernetesNamespace).List(metav1.ListOptions{LabelSelector: LABEL_WORKER_MODEL})
```
yesnault: this func returns "WorkersStartedByModel returns the number of instances of given model", but model arg is never used.

PR #2104, opened 2018-01-30
fix (worker): use API Remote Time

### LINKED ISSUE #2091: worker get queue
use timestamp of api and not timestamp or worker

this will avoid to install ntp everywhere workers are running if time is not sync with api

yesnault:
close #2091

feat (api): add process time in response headers

### DIFF
--- a/engine/api/router_util.go
+++ b/engine/api/router_util.go
@@ -74,10 +74,10 @@ func writeNoContentPostMiddleware(ctx context.Context, w http.ResponseWriter, re
 
 func writeProcessTime(w http.ResponseWriter) {
 	for k, v := range w.Header() {
-		if k == cdsclient.ResponseAPITSTimeHeader && len(v) == 1 {
+		if k == cdsclient.ResponseAPINanosecondsTimeHeader && len(v) == 1 {
 			start, err := strconv.ParseInt(v[0], 10, 64)
 			if err != nil {
-				log.Error("writeProcessTime> error on ParseInt header ResponseAPITSTimeHeader: %s", err)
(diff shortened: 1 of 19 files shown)


engine/api/router_util.go
```diff
@@ -64,10 +67,24 @@ func writeNoContentPostMiddleware(ctx context.Context, w http.ResponseWriter, re
...
 		}
 	}
+	writeProcessTime(w)
 	w.WriteHeader(http.StatusNoContent)
 	return ctx, nil
 }
 
+func writeProcessTime(w http.ResponseWriter) {
+	for k, v := range w.Header() {
+		if k == cdsclient.ResponseAPITSTimeHeader && len(v) == 1 {
```
fsamin: why not just w.Header().Get ?

sdk/cdsclient/http.go
```diff
@@ -26,6 +26,17 @@ const (
 	RequestedWithValue = "X-CDS-SDK"
 	// RequestedNameHeader is used as HTTP header
 	RequestedNameHeader = "X-Requested-Name"
+	// RequestedIfModifiedSinceHeader is used as HTTP header
+	RequestedIfModifiedSinceHeader = "If-Modified-Since"
+
+	// ResponseAPITimeHeader is used as HTTP header
+	ResponseAPITimeHeader = "X-Api-Time"
+	// ResponseAPITSTimeHeader is used as HTTP header
+	ResponseAPITSTimeHeader = "X-Api-Ts-Time"
```
fsamin: Ts ?

yesnault:
@fsamin done

PR #2116, opened 2018-01-31
feat (repositories): load files handler

fsamin:


engine/repositories/processor_loadfiles.go
```diff
@@ -0,0 +1,45 @@
+package repositories
+
+import (
+	"io/ioutil"
+
+	repo "github.com/fsamin/go-repo"
```
bnjjj: Blank Line after

engine/repositories/repositories_handlers.go
```diff
@@ -0,0 +1,52 @@
+package repositories
+
+import (
+	"context"
+	"net/http"
+
+	"github.com/gorilla/mux"
```
bnjjj: Blank Line after

PR #2118, opened 2018-02-01
Fix color on search navbar

NicoPennec:
After searching an existing project in the top search navbar, the text color should be inverted too.

The `<sui-search>` component does not allow to override his `<input>` inner child with an 'inverted` class from semantic... 

https://github.com/edcarroll/ng2-semantic-ui/blob/master/src/modules/search/components/search.ts#L19


fsamin:
Hi @NicoPennec , could you rebase with master ?

PR #2137, opened 2018-02-02
fix (vcs): get branch from bitbucket

sguiheux:
cc @arnaudmorin

### DIFF
--- a/engine/vcs/bitbucket/client_branch.go
+++ b/engine/vcs/bitbucket/client_branch.go
@@ -69,29 +69,15 @@ func (b *bitbucketClient) Branch(fullname string, filter string) (*sdk.VCSBranch
 		return nil, sdk.ErrNotFound
 	}
 
-	found := false
-	var branch sdk.VCSBranch
 	for _, b := range branches.Values {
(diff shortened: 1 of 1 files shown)


engine/vcs/bitbucket/client_branch.go
```diff
@@ -69,11 +69,29 @@ func (b *bitbucketClient) Branch(fullname string, filter string) (*sdk.VCSBranch
...
+				DisplayID:    b.DisplayID,
+				LatestCommit: b.LatestHash,
+				Default:      b.IsDefault,
+			}
+			break
+		}
+	}
+
+	if !found {
+		branch = sdk.VCSBranch{
```
arnaudmorin: so if it's not found, you will return a wrong branch?
Why not fail instead?
Failure is safer IMHO
I dont want to build a branch that looks like the one I am expecting. If I did something wrong, I just want the system to tell me fuck you

sguiheux: MIssing something yes, but not sure where, api side or vcs µservice.

PR #2138, opened 2018-02-05
fix (worker): Add insecure flag to worker cmd and overrides worker ht…

richardlt:
Here it is another pull request to deal with self-signed certs.

When starting a local hatchery for a remote CDS instance with self-signed cert, jobs are taken by hatcheries but workers won't start.
 
There is a first error occurs when registering a worker because of the worker's HTTP client. Then when the worker sends job's results there is another issue with the SDK HTTP client.

It seems that I can use the CDS_SKIP_VERIFY env var to override SDK's HTTP client but all env vars that starts with CDS or HATCHERY are not given at worker startup.

This fix works well for me, maybe this is not the best way to fix the bug. Let me know!

### DIFF
--- a/sdk/cdsclient/client.go
+++ b/sdk/cdsclient/client.go
@@ -48,7 +48,7 @@ func NewService(endpoint string, timeout time.Duration) Interface {
 }
 
 // NewWorker returns client for a worker
-func NewWorker(endpoint string, name string) Interface {
+func NewWorker(endpoint string, name string, insecureSkipVerifyTLS bool) Interface {
 	conf := Config{
 		Host:  endpoint,
 		Retry: 10,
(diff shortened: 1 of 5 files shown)


sdk/cdsclient/client.go
```diff
@@ -48,7 +48,7 @@ func NewService(endpoint string, timeout time.Duration) Interface {
 }
 
 // NewWorker returns client for a worker
-func NewWorker(endpoint string, name string) Interface {
+func NewWorker(endpoint string, name string, insecureSkipVerifyTLS bool) Interface {
```
fsamin: i would prefer to able to pass an optionnal *http.Client to the NewWorker funct

richardlt: I can change the code for a single client that overrides the two others (SDK and worker's client) if not nil. By the way there is currently a 10 seconds timeout for the worker's client but nothing for the SDK, should it really be the same client ?
Another solution would be to add the SDK override in a cmd prerun for worker like it is for CDS CLI and let the insecure flag as in NewHatchery func just below NewWorker.

fsamin: We want to remove cobra (and viper) from the worker, so don't consider the second solution :)
You are facing to the fact that we are currently in refactoring the code from the old sdk package and the new cdsclient package, so don't waste your time with the sdk package :)

PR #2140, opened 2018-02-05
chore: update venom & dep ensure with dep 0.4.1

yesnault:

warning: inexact rename detection was skipped due to too many files.
warning: you may want to set your diff.renameLimit variable to at least 2156 and retry the command.


Gopkg.lock
```diff
@@ -930,10 +931,10 @@
   revision = "d7e841db6909ae5e9eb7493f7b1ea7d7b15101e3"
 
 [[projects]]
+  branch = "master"
   name = "github.com/xanzy/go-gitlab"
   packages = ["."]
-  revision = "7bcfc0373b18ff9f02c7485a15a29053d85e8e56"
-  version = "v0.7.5"
+  revision = "e9b63f4c287702da85a1c0be96c98c8d7e98c5a5"
```
fsamin: I think i prefer use a tag why did dep change

PR #2197, opened 2018-02-09
fix(api): return when request context is done

bnjjj:


engine/api/lastupdate.go
```diff
@@ -163,6 +163,8 @@ func (b *lastUpdateBroker) ServeHTTP() Handler {
 	leave:
 		for {
 			select {
+			case <-ctx.Done():
+				return nil
```
fsamin: not sure we have to return.

i think you should do the same operations than the closeNotify case and thus just write a common `case <-w.(http.CloseNotifier).CloseNotify(), <-ctx.Done():`

PR #2199, opened 2018-02-10
refactor (api): sse - db access in lock

yesnault:


yesnault:
This PR is here to be discussed.
Two things:
 - avoid requesting database with a lock on b.mutex -> slow requests can lock all msg on api instance, then 504 return to new clients (we see a lot of them, it's probably the root cause of 504 on this endpoint).
- permissions are loaded at the beginning of SSE uuid -> we probably need something to update them when there is a permission update somewhere. 
I think this PR will not be worse in terms of performance (and probably fix at least 504, if it does not fix a memory leak).

bnjjj:
Maybe https://github.com/ovh/cds/pull/2199/files#diff-9d30c6aa4afc5f8149d08483db3477e5R77 set b.clients to nil should be more efficient too ?

PR #2251, opened 2018-02-17
fix(api): return error if two model are selected in requirements

bnjjj:


engine/api/action/action.go
```diff
@@ -310,12 +310,20 @@ func UpdateActionDB(db gorp.SqlExecutor, a *sdk.Action, userID int64) error {
...
 				return sdk.ErrInvalidJobRequirement
 			}
 		}
+		if a.Requirements[i].Type == sdk.ModelRequirement {
+			nbModelReq++
+		}
+	}
+
+	if nbModelReq > 0 {
+		return sdk.ErrInvalidJobRequirement
```
yesnault: it could be more useful for user if he see: 
"you can't select more than one model in job requirement"
instead of "Invalid job requirement"

sdk/error.go
```diff
@@ -381,6 +383,7 @@ var errorsFrench = map[int]string{
 	ErrDownloadDoesNotExist.ID:                  "Le fichier n'existe pas",
 	ErrTokenNotFound.ID:                         "Le token n'existe pas",
 	ErrWorkflowNotificationNodeRef.ID:           "Une référence de noeud de workflow est invalide dans vos notifications (si vous souhaitez supprimer un pipeline vérifiez qu'il ne soit plus référencé dans la liste de vos notifications)",
+	ErrInvalidJobRequirementDuplicateModel.ID:   "Pré-requis de job invalide: vous ne pouvez pas séléctionnez plusieurs modèles de worker",
```
fsamin: invalides

sdk/error.go
```diff
@@ -258,6 +259,7 @@ var errorsAmericanEnglish = map[int]string{
 	ErrDownloadDoesNotExist.ID:                  "File does not exist",
 	ErrTokenNotFound.ID:                         "Token does not exist",
 	ErrWorkflowNotificationNodeRef.ID:           "An invalid workflow node reference has been found, if you want to delete a pipeline from your workflow check if this pipeline isn't referenced in your notifications list",
+	ErrInvalidJobRequirementDuplicateModel.ID:   "Invalid job requirement: you can't select multiple worker models",
```
fsamin: requirements

sguiheux:
maybe add something on UI too?

PR #2255, opened 2018-02-19
feat: migration from key variable to sshkey

sguiheux:


engine/api/migrate/key.go
```diff
@@ -0,0 +1,277 @@
...
+
+	// Project migration
+	ids, errQ := loadProjectIDs(db)
+	if errQ != nil {
+		return
+	}
+
+	for _, id := range ids {
+		if err := migrateProject(db, id, store, u); err != nil {
+			continue
```
bnjjj: useless. Do we need some warning logs ?

engine/api/migrate/key.go
```diff
@@ -0,0 +1,277 @@
...
+	}
+
+	// Application migration
+	appIds, errA := loadApplicationIDs(db)
+	if errA != nil {
+		return
+	}
+	for _, id := range appIds {
+		if err := migrateApplication(db, id, store, u); err != nil {
+			continue
```
bnjjj: useless

engine/api/migrate/key.go
```diff
@@ -0,0 +1,277 @@
...
+	}
+
+	// Environment migration
+	envIds, errE := loadEnvironmentIDs(db)
+	if errE != nil {
+		return
+	}
+	for _, id := range envIds {
+		if err := migrateEnvironment(db, id); err != nil {
+			continue
```
bnjjj: useless

engine/api/migrate/key.go
```diff
@@ -0,0 +1,277 @@
...
+	tx, errT := db.Begin()
+	if errT != nil {
+		log.Warning("migrateApplication> Cannot start transaction: %s", errT)
+		return errT
+	}
+
+	app, errA := application.LoadAndLockByID(tx, store, appID, u, application.LoadOptions.WithVariablesWithClearPassword, application.LoadOptions.WithKeys)
+	if errA != nil {
+		log.Warning("migrateApplication> Cannot load application %d: %s", appID, errA)
+	}
```
bnjjj: missing return ? because if there is an error the app after will be nil, and it will panic

PR #2282, opened 2018-02-22
fix (engine): use execGroups - worker checks wm groups

yesnault:
This will avoid a worker model not shared.infra to run a job not attached to his group


sguiheux:
I agree with @fsamin.  When Hatchery/Worker do a GetQueue, only job that can be executed must be returned. 
So checking the group with requirement is useless

yesnault:
I agree: 
- keep execGroups
- impact getQueue to return only good jobs
- remove checkreq perm in worker
- check perm in take

engine/api/workflow/execute_node_run.go
```diff
@@ -362,6 +370,57 @@ func addJobsToQueue(db gorp.SqlExecutor, stage *sdk.Stage, run *sdk.WorkflowNode
 	return nil
 }
 
+func getJobExecutablesGroups(db gorp.SqlExecutor, run *sdk.WorkflowNodeRun) ([]sdk.Group, error) {
```
fsamin: do not return groups, add them to the workflowNodeRun

fsamin: Should be nice to have unit tests of this huge sql request

engine/worker/run.go
```diff
@@ -91,9 +91,9 @@ func (w *currentWorker) processActionVariables(a *sdk.Action, parent *sdk.Action
 }
 
 func (w *currentWorker) startAction(ctx context.Context, a *sdk.Action, buildID int64, params *[]sdk.Parameter, stepOrder int, stepName string) sdk.Result {
-	log.Debug("startAction> Begin %p", ctx)
+	log.Debug("startAction> Begin %s", a.ID)
```
fsamin: the logs was here to follow the context propagation; now you can remove it

engine/sql/082_wnrj_groups.sql
```diff
@@ -0,0 +1,6 @@
+-- +migrate Up
+ALTER TABLE workflow_node_run_job ADD COLUMN exec_groups JSONB;
```
sguiheux: rename fto 083

engine/api/workflow/execute_node_run.go
yesnault: no groups attr on WorkflowNodeRun, see. https://github.com/ovh/cds/blob/master/sdk/workflow_run.go#L99

engine/api/workflow/dao_node_run_job.go
```diff
@@ -71,7 +71,7 @@ func CountNodeJobRunQueue(db gorp.SqlExecutor, store cache.Store, groupsID []int
 }
 
 // LoadNodeJobRunQueue load all workflow_node_run_job accessible
-func LoadNodeJobRunQueue(db gorp.SqlExecutor, store cache.Store, groupsID []int64, since *time.Time, statuses ...string) ([]sdk.WorkflowNodeJobRun, error) {
+func LoadNodeJobRunQueue(db gorp.SqlExecutor, store cache.Store, isUser bool, groupsID []int64, since *time.Time, statuses ...string) ([]sdk.WorkflowNodeJobRun, error) {
```
fsamin: is think it's more consistent to pass the user than a boolean

engine/api/workflow/dao_node_run_job.go
```diff
@@ -90,10 +90,29 @@ func LoadNodeJobRunQueue(db gorp.SqlExecutor, store cache.Store, groupsID []int6
 
 	jobs := make([]sdk.WorkflowNodeJobRun, len(sqlJobs))
 	for i := range sqlJobs {
-		getHatcheryInfo(store, &sqlJobs[i])
 		if err := sqlJobs[i].PostGet(db); err != nil {
 			return nil, sdk.WrapError(err, "workflow.LoadNodeJobRun> Unable to load job runs (PostGet)")
 		}
+		var toadd bool
```
fsamin: this var name is pretty hard to understand this sounds like toad and i don't understand what we have to add

engine/api/workflow/execute_node_run.go
fsamin: ?

fsamin: ok

PR #2285, opened 2018-02-23
fix (api): no check for removing group on default env

yesnault:


yesnault:
Root cause: noEnv is in pipeline_trigger, with src_environment_id = 1 for pipeline type build.

**before merging: check if no 082 migrate script on master **

PR #2288, opened 2018-02-23
fix(api): flush lastupdate buffer by default

bnjjj:


bnjjj:
ready to merge :)

PR #2294, opened 2018-02-26
fix(api): workflow as code header

fsamin:


sdk/cdsclient/client_import.go
```diff
@@ -7,6 +7,8 @@ import (
 	"io"
 	"net/http"
 
+	"github.com/ovh/cds/sdk"
```
sguiheux: good try but no

PR #2328, opened 2018-02-28
fix(api,vcs): inconsistent statuses

fsamin:


engine/vcs/github/client_status.go
```diff
@@ -184,7 +185,8 @@ func processEventWorkflowNodeRun(event sdk.Event, githubURL string, disabledStat
 		data.urlPipeline = ""
 	}
 
-	data.desc = sdk.VCSCommitStatusDescription(eventNR)
+	data.context = sdk.VCSCommitStatusDescription(eventNR)
```
bnjjj: Is the name of the function still valid ?

fsamin: Yes, it is. For gitlab and bitbucket, it is the description; for github we have description (human readable) and context (more technical stuff)

PR #2339, opened 2018-03-01
fix (api): upload artifact - compare obj

yesnault:


engine/api/workflow_queue_artifact.go
```diff
@@ -199,7 +200,7 @@ func (api *API) postWorkflowJobArtifactWithTempURLCallbackHandler() Handler {
 			return sdk.WrapError(sdk.ErrNotFound, "postWorkflowJobArtifactWithTempURLCallbackHandler> Unable to find artifact, key:%s", cacheKey)
 		}
 
-		if art != cachedArt {
+		if !cmp.Equal(art, cachedArt) {
```
fsamin: What was the difference between the object ?
The struct in only composed of comparable types, so we don't need reflection to compare it

Eventually if the comparison is a bit specific implement an Equal function on the type to avoid a new dependency and reflection

PR #2360, opened 2018-03-08
fix(api): return error when branch is deleted from repository during …

sguiheux:
…workflow

### DIFF
--- a/engine/api/workflow/process.go
+++ b/engine/api/workflow/process.go
@@ -497,12 +497,14 @@ func processWorkflowNodeRun(dbCopy *gorp.DbMap, db gorp.SqlExecutor, store cache
 		isRoot = true
 	}
 
-	vcsInfos, errVcs := getVCSInfos(db, store, p, gitValues, n, run, !isRoot)
+	var previousGitRepo string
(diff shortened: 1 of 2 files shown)


engine/api/workflow/process.go
```diff
@@ -497,12 +497,14 @@ func processWorkflowNodeRun(dbCopy *gorp.DbMap, db gorp.SqlExecutor, store cache
...
+	if pGitRepo != nil {
+		previousGitRepo = pGitRepo.Value
+	}
+	vcsInfos, errVcs := getVCSInfos(db, store, p, gitValues, n, run, !isRoot, previousGitRepo)
 	if errVcs != nil {
-		if isRoot {
-			return false, errVcs
-		}
-		log.Error("processWorkflowNodeRun> Cannot get VCSInfos")
+		return false, sdk.WrapError(errVcs, "processWorkflowNodeRun> Cannot get VCSInfos")
```
fsamin: i think return an error will raison an error through the http handler; in this case we should set the workflow in failed status

sguiheux: no :)

PR #2407, opened 2018-03-19
 fix (api): avoid sending notif on default group

### LINKED ISSUE #2275: fix (api): avoid sending jabb notif on default group

yesnault:
close #2275

### DIFF
--- a/engine/api/group.go
+++ b/engine/api/group.go
@@ -195,6 +195,7 @@ func (api *API) getGroupsHandler() Handler {
 		var err error
 
 		public := r.FormValue("withPublic")
+		withoutDefault := r.FormValue("withoutDefault")
 		if getUser(ctx).Admin {
 			groups, err = group.LoadGroups(api.mustDB())
 		} else {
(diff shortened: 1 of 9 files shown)


engine/api/group.go
```diff
@@ -195,6 +195,7 @@ func (api *API) getGroupsHandler() Handler {
 		var err error
 
 		public := r.FormValue("withPublic")
+		withoutDefault := r.FormValue("withoutDefault")
```
bnjjj: use `FormBool(r, "withoutDefault")` is better I think

ui/src/app/service/group/group.service.ts
```diff
@@ -31,6 +31,14 @@ export class GroupService {
         return this._http.get<Group[]>('/group');
     }
 
+    /**
+     * Get all groups that the user can access, without the default group.
+     * @returns {Observable<Group[]>}
+     */
+    getGroupsWithoutDefault(): Observable<Group[]> {
```
bnjjj: why not pass an optional parameters to `getGroups(withoutDefault?: boolean)` and then use http query parameters of angular

PR #2408, opened 2018-03-19
feat (engine): --is-uptodate flag on engine update

### LINKED ISSUE #2369: engine update [--is-uptodate] [--api...]
returns true or false

check if current binary is up to date with remote

yesnault:
close #2369

cc @henyxia

### DIFF
--- a/engine/main.go
+++ b/engine/main.go
@@ -54,6 +54,7 @@ func init() {
 	//Update  command
 	mainCmd.AddCommand(updateCmd)
 	updateCmd.Flags().BoolVar(&updateFromGithub, "from-github", false, "Update binary from latest github release")
+	updateCmd.Flags().BoolVar(&isUptodate, "is-uptodate", false, "return true or false is engine binary is uptodate or not")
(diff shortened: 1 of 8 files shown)


engine/main.go
```diff
@@ -54,6 +54,7 @@ func init() {
 	//Update  command
 	mainCmd.AddCommand(updateCmd)
 	updateCmd.Flags().BoolVar(&updateFromGithub, "from-github", false, "Update binary from latest github release")
+	updateCmd.Flags().BoolVar(&isUptodate, "is-uptodate", false, "return true or false is engine binary is uptodate or not")
```
bnjjj: the term true or false is not correct I think

engine/update.go
```diff
@@ -13,14 +15,14 @@ import (
...
 
 var updateFromGithub bool
+var isUptodate bool
 var updateURLAPI string
 
 var updateCmd = &cobra.Command{
 	Use:     "update",
 	Short:   "Update engine binary",
-	Example: "engine update --from-github",
+	Example: "engine update --from-github --is-uptodate",
```
fsamin: i would prefer a new command with a nice exit code

PR #2413, opened 2018-03-19
fix(api): keep the payload git from sources

bnjjj:


engine/api/workflow/process.go
```diff
@@ -485,7 +485,7 @@ func processWorkflowNodeRun(dbCopy *gorp.DbMap, db gorp.SqlExecutor, store cache
 
 	//Parse job params to get the VCS infos
 	gitValues := map[string]string{}
-	for _, param := range jobParams {
+	for _, param := range run.BuildParameters {
```
fsamin: i'm really not sure about this, it's a very sensitive part of code

PR #2427, opened 2018-03-21
Refactor gorp json

fsamin:
[PROPOSAL]

i try to manage automatically with gorp the json field of our entities.
Here it's done with sdk.Application. 

For the moment, we have to call manually the function in the DAO, but we can remove the `PostInsert` and `PostUpdate` hooks.

Next step is to remove to be able to remove the `PostGet` hooks.

What do you think ? would it be fine for you ?

FYI: the technical part of the code is in the commit 8e962b4, usage of the feature on an entity is in commit 0ce26ab

### DIFF
--- a/engine/api/application/dao.go
+++ b/engine/api/application/dao.go
@@ -10,6 +10,7 @@ import (
 	"github.com/lib/pq"
 
 	"github.com/ovh/cds/engine/api/cache"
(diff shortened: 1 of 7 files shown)


fsamin:
finally is think if will override a gorp.Dialect as explained here: https://github.com/go-gorp/gorp/issues/254#issuecomment-248733253

PR #2463, opened 2018-03-24
feat(ui): add feature to import pipeline with yaml from UI

bnjjj:
1. Description
Add ability to import pipeline with yaml from the UI
1. Related issues
1. About tests
1. Mentions

@ovh/cds


yesnault:
It will be great to add a example of yaml on the pipeline creation page.

capture d ecran 2018-03-25 a 20 04 54

Example:

```yaml
name: MyPipeline
stages:
  1|Build:
    jobs:
      MyJob:
        steps:
        - script: echo I'm the firt step
        - script: echo I'm the second step
```
In addition, is it possible to add a way to upload an existing file?

fsamin:
this is an old-school yaml pipeline !

ui/src/app/views/pipeline/add/pipeline.add.component.ts
```diff
@@ -53,4 +64,18 @@ export class PipelineAddComponent {
...
 
     }
+
+    goToProject(): void {
+        this._router.navigate(['/project', this.project.key], {queryParams: {tab: 'pipelines'}});
+    }
+
+    importPipeline() {
+        this.loadingCreate = true;
+        this._pipStore.importPipeline(this.project.key, this.pipToImport)
```
sguiheux: Possible to add a popup warning when a pipeline with the same name exist?

bnjjj: forceUpdate deleted

PR #2474, opened 2018-03-27
fix(ui): navbar switch + navbar search

sguiheux:
1. Description
Cannot switch workflow via navbar
1. Related issues
1. About tests
1. Mentions

@ovh/cds

### DIFF
--- a/ui/src/app/views/navbar/navbar.component.ts
+++ b/ui/src/app/views/navbar/navbar.component.ts
@@ -115,14 +115,18 @@ export class NavbarComponent implements OnInit, AfterViewInit {
             if (workflows) {
                 this.navRecentWorkflows = workflows;
                 this.listWorkflows = workflows;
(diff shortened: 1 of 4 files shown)


ui/src/app/views/navbar/navbar.component.ts
```diff
@@ -115,14 +115,18 @@ export class NavbarComponent implements OnInit, AfterViewInit {
             if (workflows) {
                 this.navRecentWorkflows = workflows;
                 this.listWorkflows = workflows;
-                this.recentItems = workflows.toArray()
-                    .map((w) => ({
+                this.recentItems = new Array<NavbarSearchItem>();
+                workflows.toArray().forEach(w => {
+                    if (!w.name) {
+                        return;
```
bnjjj: return doesn't work inside forEach, use if (w.name) instead

PR #2478, opened 2018-03-27
fix (engine): scheduler enqueue

yesnault:
1. Description
A new goroutine dedicated for enqueue scheduler, 30s
Existing goroutine (120s retry) works only on very old hooks or error

1. Related issues
n/a
1. About tests
Manual done
1. Mentions

@ovh/cds

### DIFF
--- a/engine/hooks/scheduler.go
+++ b/engine/hooks/scheduler.go
@@ -60,12 +60,19 @@ func (s *Service) retryTaskExecutionsRoutine(c context.Context) error {
 					continue
 				}
 				for _, e := range execs {
-
 					if e.Status == TaskExecutionDoing {
 						continue
 					}
+
+					if (e.Type == TypeScheduler || e.Type == TypeRepoPoller) && e.ProcessingTimestamp == 0 && e.Timestamp <= time.Now().UnixNano() {
+						log.Warning("Enqueing scheduler/repoPoller %s %d/%d  %s", e.UUID, e.NbErrors, s.Cfg.RetryError, e.LastError)
+						s.Dao.EnqueueTaskExecution(&e)
+						continue
+					}
+
+					// old hooks
 					if e.ProcessingTimestamp == 0 && e.Timestamp < time.Now().Add(-1*time.Hour).UnixNano() {
-						log.Warning("Enqueing %s %d/%d  %s", e.UUID, e.NbErrors, s.Cfg.RetryError, e.LastError)
+						log.Warning("Enqueing very old hooks %s %d/%d  %s", e.UUID, e.NbErrors, s.Cfg.RetryError, e.LastError)
 						s.Dao.EnqueueTaskExecution(&e)
 					}
 					if e.NbErrors < s.Cfg.RetryError && e.LastError != "" {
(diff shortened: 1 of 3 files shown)


engine/hooks/scheduler.go
```diff
@@ -79,6 +86,41 @@ func (s *Service) retryTaskExecutionsRoutine(c context.Context) error {
 	}
 }
 
+// Every x seconds, the scehduler try to relaunch all tasks which have never been processed, or in error
+func (s *Service) enqueueScheduledTaskExecutionsRoutine(c context.Context) error {
+	tick := time.NewTicker(time.Duration(30) * time.Second)
+	for {
```
fsamin: i think it's better to do a defer tick.Stop() than tick.Stop before the return

engine/hooks/scheduler.go
```diff
@@ -79,6 +86,41 @@ func (s *Service) retryTaskExecutionsRoutine(c context.Context) error {
...
+					log.Error("Hooks> enqueueScheduledTaskExecutionsRoutine > Unable to find all task executions (%s): %v", t.UUID, err)
+					continue
+				}
+				for _, e := range execs {
+					if e.Status == TaskExecutionScheduled && e.ProcessingTimestamp == 0 && e.Timestamp <= time.Now().UnixNano() {
+						// update status before enqueue
+						// this will avoid to re-enqueue the same scheduled task execution if the dequeue take more than 30s (ticker of this goroutine)
+						e.Status = ""
+						s.Dao.SaveTaskExecution(&e)
+						log.Warning("Enqueing scheduler/repoPoller %s", e.UUID)
```
fsamin: scheduler/repoPoller ?

engine/hooks/scheduler.go
```diff
@@ -79,6 +86,41 @@ func (s *Service) retryTaskExecutionsRoutine(c context.Context) error {
 	}
 }
 
+// Every x seconds, the scehduler try to relaunch all tasks which have never been processed, or in error
```
fsamin: typo

engine/hooks/scheduler.go
```diff
@@ -79,6 +86,41 @@ func (s *Service) retryTaskExecutionsRoutine(c context.Context) error {
...
+func (s *Service) enqueueScheduledTaskExecutionsRoutine(c context.Context) error {
```
fsamin: Why not enqueueNeverFiredoOErrrorTaskExecutionRoutine ?

yesnault:
root cause: func retryTaskExecutionsRoutine works on task on error and task with older than 1hour (previous fix).
But, this same func was used to launch scheduled tasks and repo poller task too.

So, instead of using the same func to launch error, older and "normal" scheduler & repo poller task, we create another func enqueueScheduledTaskExecutionsRoutine to work only on scheduled & repo poller task to put them in queue.

To avoid re-enqueue them, we use a new status TaskExecutionScheduled, update it to "" just before calling s.Dao.EnqueueTaskExecution(&e)

With that fix, the scheduled task is run "in time", the routine enqueueScheduledTaskExecutionsRoutine has a 30s ticker, instead of 120s for retry (prod value).

PR #2564, opened 2018-04-11
fix (vcs): take care of RateLimitReset

yesnault:

### DIFF
--- a/engine/vcs/github/client_rate_limit.go
+++ b/engine/vcs/github/client_rate_limit.go
@@ -2,10 +2,19 @@ package github
 
 import (
 	"encoding/json"
+	"time"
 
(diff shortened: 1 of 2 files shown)


engine/vcs/github/client_rate_limit.go
```diff
@@ -2,11 +2,20 @@ package github
...
 import (
 	"encoding/json"
+	"time"
 
 	"github.com/ovh/cds/sdk"
 	"github.com/ovh/cds/sdk/log"
 )
 
+func isRateLimitReached() bool {
+	if RateLimitReset > 0 && RateLimitReset < int(time.Now().Unix()) {
```
fsamin: asking for a friend: should we put a mutex in there ?

PR #2572, opened 2018-04-15
feat (engine): status on all services, api too

yesnault:
hatcheries -> services (old heartbeat on hatchery is keep for now)
api -> service too
status is stored in service table 

so, cdsctl admin services status [-t]

with type :
-t, --type string     Filter service by type: api, hatchery, hook, repository, vcs

### DIFF
--- a/engine/api/services.go
+++ b/engine/api/services.go
@@ -110,7 +110,8 @@ func (api *API) serviceAPIHeartbeat(c context.Context) {
 			//Try to find the service, and keep; else generate a new one
 			oldSrv, errOldSrv := repo.FindByName(srv.Name)
 			if errOldSrv != nil && errOldSrv != sdk.ErrNotFound {
-				log.Error("serviceAPIHeartbeat:%v", errOldSrv)
+				log.Error("serviceAPIHeartbeat> Unable to find by name:%v", errOldSrv)
+				repo.Rollback()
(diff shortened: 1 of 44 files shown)


engine/api/services.go
```diff
@@ -73,3 +75,65 @@ func (api *API) postServiceRegisterHandler() Handler {
...
+				MonitoringStatus: api.Status(),
+				Hash:             string(hash),
+				LastHeartbeat:    time.Now(),
+			}
+
+			//Try to find the service, and keep; else generate a new one
+			oldSrv, errOldSrv := repo.FindByName(srv.Name)
+			if errOldSrv != nil && errOldSrv != sdk.ErrNotFound {
+				log.Error("serviceAPIHeartbeat:%v", errOldSrv)
+				continue
```
bnjjj: missing repo.Rollback() ?

engine/hatchery/marathon/marathon.go
```diff
@@ -14,8 +14,11 @@ import (
 
 	"github.com/facebookgo/httpcontrol"
 	"github.com/gambol99/go-marathon"
+	"github.com/gorilla/mux"
 	"github.com/ovh/cds/sdk/namesgenerator"
```
bnjjj: move this to the good import block

PR #2581, opened 2018-04-17
fix (hatchery): call status on not fully initialized service

yesnault:


yesnault:
this will avoid nil pointer when hatchery is not fully initialized.

PR #2611, opened 2018-04-26
fix(api): add better log for application import

bnjjj:

1. Description
1. Related issues
1. About tests
1. Mentions

@ovh/cds

### DIFF
--- a/engine/api/application_import.go
+++ b/engine/api/application_import.go
@@ -15,6 +15,7 @@ import (
 	"github.com/ovh/cds/engine/api/project"
 	"github.com/ovh/cds/sdk"
 	"github.com/ovh/cds/sdk/exportentities"
+	"github.com/ovh/cds/sdk/log"
 )
 
 func (api *API) postApplicationImportHandler() Handler {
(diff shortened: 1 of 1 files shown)


engine/api/application_import.go
```diff
@@ -68,7 +69,8 @@ func (api *API) postApplicationImportHandler() Handler {
 		if globalError != nil {
 			myError, ok := globalError.(sdk.Error)
 			if ok {
-				return WriteJSON(w, msgListString, myError.Status)
+				log.Warning("postApplicationImportHandler> Unable to import application %s : %s", eapp.Name, myError.String())
```
fsamin: %v  -> myError

engine/api/application_import.go
```diff
@@ -68,7 +69,8 @@ func (api *API) postApplicationImportHandler() Handler {
...
+				return WriteJSON(w, myError.String(), myError.Status)
```
fsamin: humm no
we have to send msgListString because there can be useful message for the user.
Why not add the the error at the end of the list (with the right translation) 

btw: return the error should be enough because does a WriteJSON(...) of the error

PR #2612, opened 2018-04-26
fix(api): add ctx with timeout of 2 minutes to stop infinite lock (quick fix)

bnjjj:

1. Description
1. Related issues
1. About tests
1. Mentions

@ovh/cds

### DIFF
--- a/engine/api/workflow_queue.go
+++ b/engine/api/workflow_queue.go
@@ -349,7 +349,8 @@ func (api *API) postWorkflowJobResultHandler() Handler {
 		if err := UnmarshalBody(r, &res); err != nil {
 			return sdk.WrapError(err, "postWorkflowJobResultHandler> cannot unmarshal request")
 		}
-		dbWithCtx := api.mustDBWithCtx(ctx)
+		customCtx, _ := context.WithTimeout(context.Background(), 90*time.Second)
+		dbWithCtx := api.mustDBWithCtx(customCtx)
 
 		proj, errP := project.LoadProjectByNodeJobRunID(dbWithCtx, api.Cache, id, getUser(ctx), project.LoadOptions.WithVariables)
 		if errP != nil {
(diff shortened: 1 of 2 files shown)


bnjjj:
It's a temporary quick fix, the root cause must be fixed in another way

engine/api/workflow_queue.go
```diff
@@ -349,7 +349,8 @@ func (api *API) postWorkflowJobResultHandler() Handler {
 		if err := UnmarshalBody(r, &res); err != nil {
 			return sdk.WrapError(err, "postWorkflowJobResultHandler> cannot unmarshal request")
 		}
-		dbWithCtx := api.mustDBWithCtx(ctx)
+		customCtx, _ := context.WithTimeout(context.Background(), 90*time.Second)
```
fsamin: defer call the cancel func, to avoid having a timer pending for a while

engine/api/workflow_queue.go
```diff
@@ -375,9 +376,10 @@ func (api *API) postWorkflowJobResultHandler() Handler {
 		if err != nil {
 			return err
 		}
+		db := api.mustDB()
 
-		workflow.ResyncNodeRunsWithCommits(api.mustDB(), api.Cache, proj, workflowNodeRuns)
-		go workflow.SendEvent(api.mustDB(), workflowRuns, workflowNodeRuns, workflowNodeJobRuns, proj.Key)
+		workflow.ResyncNodeRunsWithCommits(db, api.Cache, proj, workflowNodeRuns)
+		go workflow.SendEvent(db, workflowRuns, workflowNodeRuns, workflowNodeJobRuns, proj.Key)
```
fsamin: why is this a goroutine ?

engine/api/workflow_queue.go
bnjjj: Nope I can't because it will be almost the same than create a db connection with request context

engine/api/workflow_queue.go
bnjjj: Maybe because it could be asynchronously done, it's not useful for the return of the handler

engine/api/workflow_queue.go
fsamin: all of this is so sad

bnjjj: need to be fixed as discussed

PR #2620, opened 2018-05-01
feat (engine): openstack auth v3

yesnault:


fsamin:
is the hatchery also concerned ?

yesnault:
@fsamin yes, I'll add it in this PR.

PR #2643, opened 2018-05-08
feat (ui): info on tag field with cds workflows

yesnault:
This is temporary, tag field have to be removed
from action when old workflows will be removed from CDS Engine.

But this will avoid misunderstanding for users using old and new workflows.


yesnault:
if you have a better solution, I'm interested

PR #2650, opened 2018-05-09
fix: use redis pubsub for sse events

sguiheux:
1. Description
1. Related issues
1. About tests
1. Mentions

@ovh/cds

### DIFF
--- a/engine/api/event/publish.go
+++ b/engine/api/event/publish.go
@@ -7,9 +7,11 @@ import (
 	"github.com/fatih/structs"
 	"github.com/go-gorp/gorp"
 
+	"encoding/json"
 	"github.com/ovh/cds/engine/api/cache"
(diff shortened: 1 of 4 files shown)


engine/api/event/publish.go
```diff
@@ -19,6 +21,12 @@ func publishEvent(e sdk.Event) {
 	Cache.Enqueue("events", e)
 	// send to cache for cds repositories manager
 	Cache.Enqueue("events_repositoriesmanager", e)
+
+	b, err := json.Marshal(e)
+	if err != nil {
+		log.Warning("publishEvent> Cannot marshal event %+v", e)
+	}
+	Cache.Publish("events_pubsub", string(b))
```
yesnault: If there is an error on marshal, the publish won't work (or publish 'nil' here)

PR #2692, opened 2018-05-17
feat (api): multi-repo

yesnault:


engine/api/workflow/process.go
```diff
@@ -498,12 +499,17 @@ func processWorkflowNodeRun(dbCopy *gorp.DbMap, db gorp.SqlExecutor, store cache
 	}
 
 	//Parse job params to get the VCS infos
+	var currentGitValuesAlreadyKnowned bool
```
bnjjj: I think knowned doesn't exist. prefer known

sdk/workflow_run.go
```diff
@@ -79,6 +79,20 @@ func (r *WorkflowRun) Tag(tag, value string) bool {
 	return false
 }
 
+// TagExists return true if tag already exits
+func (r *WorkflowRun) TagExists(tag, value string) bool {
+	if value == "" {
+		return false
+	}
+	var found bool
```
bnjjj: found is not used ? Am I wrong ?

PR #2700, opened 2018-05-18
fix(api): return empty json when you return nil, linked to firefox bug

bnjjj:

1. Description
1. Related issues
1. About tests
1. Mentions

@ovh/cds


bnjjj:
@fsamin I dont really understand how do you want to do. Because we don't know at this place if we have written something in the writer. Am I wrong ? Moreover sometimes we just don't return anything but it's not json content type

PR #2729, opened 2018-05-22
fix (api): avoid 401 on ui for vcs

yesnault:
@ovh/cds

### DIFF
--- a/engine/api/repositories_manager.go
+++ b/engine/api/repositories_manager.go
@@ -325,7 +325,7 @@ func (api *API) getReposFromRepositoriesManagerHandler() Handler {
 			api.Cache.SetWithTTL(cacheKey, repos, 0)
 			if errRepos != nil {
 				// as client.Repos can return a 401 error, we avoid here to return 401 on UI too.
-				return sdk.WrapError(sdk.ErrUnknownError, "getReposFromRepositoriesManagerHandler> Cannot get repos: %v", errRepos)
+				return sdk.WrapError(sdk.ErrNoReposManagerClientAuth, "getReposFromRepositoriesManagerHandler> Cannot get repos: %v", errRepos)
 			}
 		}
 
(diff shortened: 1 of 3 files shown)


engine/api/repositories_manager.go
```diff
@@ -321,14 +320,15 @@ func (api *API) getReposFromRepositoriesManagerHandler() Handler {
 
 		var repos []sdk.VCSRepo
 		if !api.Cache.Get(cacheKey, &repos) || len(repos) == 0 {
-			log.Debug("getReposFromRepositoriesManagerHandler> loading from Stash")
-			repos, err = client.Repos()
+			var errRepos error
```
fsamin: BTW i would prefer renaming the err in the parent scope rather than this limited scope

engine/api/workflow_run.go
```diff
@@ -485,7 +485,8 @@ func (api *API) getWorkflowCommitsHandler() Handler {
 
 		commits, _, errC := workflow.GetNodeRunBuildCommits(api.mustDB(), api.Cache, proj, wf, nodeName, wfRun.Number, wfNodeRun, nodeCtx.Application, nodeCtx.Environment)
 		if errC != nil {
-			return sdk.WrapError(errC, "getWorkflowCommitsHandler> Unable to load commits")
+			// as workflow.GetNodeRunBuildCommits can return a 401 error, we avoid here to return 401 on UI too
+			return sdk.WrapError(sdk.ErrNoReposManagerClientAuth, "getWorkflowCommitsHandler> Unable to load commits: %v", errC)
```
fsamin: Is a Forbidden status code okay ? (asking for a friend)

engine/api/repositories_manager.go
yesnault: done

engine/api/workflow_run.go
yesnault: yes: 	ErrNoReposManagerClientAuth               = Error{ID: 44, Status: http.StatusForbidden}

PR #2737, opened 2018-05-23
feat(api,ui) public platforms

fsamin:
1. Description
1. Related issues
1. About tests
1. Mentions

@ovh/cds


ui/src/app/views/project/show/platforms/form/platform.form.component.ts
```diff
@@ -24,6 +24,9 @@ export class ProjectPlatformFormComponent {
                 private _toast: ToastService, private _translate: TranslateService) {
         this.newPlatform = new ProjectPlatform();
         this._platformService.getPlatformModels().pipe(first()).subscribe(platfs => {
+            platfs.filter(pf => {
```
bnjjj: useless, filter is not mutable. prefer use
`platfs = platfs.filter...`

Issue #2540, opened 2018-04-06
tests: get applications and workflows performance


fsamin:
Ensure that the list of applications and workflows can be displayed in the 1s-2s range even with 300+ applications/workflows.

PR #2792, opened 2018-05-31
test(api): add benchmark for workflows

bnjjj:
1. Description
1. Related issues
Partially #2540
1. About tests
1. Mentions

@ovh/cds

### DIFF
--- a/engine/api/workflow_test.go
+++ b/engine/api/workflow_test.go
@@ -348,3 +348,75 @@ func Test_deleteWorkflowHandler(t *testing.T) {
 	router.Mux.ServeHTTP(w, req)
 	assert.Equal(t, 200, w.Code)
 }
+
+func BenchmarkGetWorkflows(b *testing.B) {
+	// Init database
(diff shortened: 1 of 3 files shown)


engine/api/workflow_test.go
```diff
@@ -348,3 +348,142 @@ func Test_deleteWorkflowHandler(t *testing.T) {
...
+	}
+	uri := router.GetRoute("GET", api.getWorkflowsHandler, vars)
+	test.NotEmpty(t, uri)
+
+	res := testing.Benchmark(func(b *testing.B) {
+		b.ResetTimer()
+
+		for n := 0; n < b.N; n++ {
+			b.StopTimer()
+			req := assets.NewAuthentifiedRequest(t, u, pass, "GET", uri, vars)
```
fsamin: you should make this request with a non admin user

bnjjj: you're right

bnjjj: done

PR #2793, opened 2018-05-31
fix(api): add history length on workflow as code

### LINKED ISSUE #2527: missing history length on workflow as code

bnjjj:

1. Description
1. Related issues
close #2527
1. About tests
1. Mentions

@ovh/cds


sdk/exportentities/workflow.go
```diff
@@ -77,6 +78,10 @@ func NewWorkflow(w sdk.Workflow, withPermission bool) (Workflow, error) {
 		}
 	}
 
+	if w.HistoryLength > 0 {
```
fsamin: and if != default value (20)

bnjjj: done

Issue #1625, opened 2017-12-04
feat (ui): add admin view in CDS for administrator


bnjjj:

yesnault:
Admin status -> https://github.com/ovh/cds/pull/1740
TODO -> 
- workers and action to change disable a worker
- worker model with direct link to worker model settings
- Queue : Workflow Run Job Waiting, and action to stop one or many workflow
- Current Workflow Run Job Building
- Overwall Status page: NB Queue, Nb Building, Resume Status (red if 1 AL on status page)

PR #2797, opened 2018-05-31
feat(ui): add queue view in UI admin section

bnjjj:

1. Description
1. Related issues
partially #1625
1. About tests
1. Mentions

@ovh/cds


yesnault:
coool !

This page should be accessible to user too, no?

admin should see everything
user should see only their jobs (as cdsctl)

bnjjj:
@yesnault Of course but we need to find where display (in the settings menu ? or in the "admin" menu ?) and if you find a better wording too :)

yesnault:
oh yes, it's just a wording.

admin menu -> it's for CDS Admin only (I think).  This menu should not be display to users
Settings -> Advanced? (we will confirm)

yesnault:
So, on the navbar:
 - one icon (<i class="signal icon"></i>) on the right of icon of settings, with url: /queue
 - another icon ( <i class="heartbeat icon"></i> ) for access to status page (actually on /admin/status -> /status) 

and "status" could be removed from "admin" menu

bnjjj:
2 more icons will be too much in the navbar. It's too ugly. At the moment I think it's better to put this link in the settings menu after the plain line separation

yesnault:
@bnjjj less icons with that: 

capture d ecran 2018-06-03 a 11 15 29
capture d ecran 2018-06-03 a 11 26 38

Should be ok, no ? (the language switcher should be only in profile?)

PR #2814, opened 2018-06-05
fix (api): check error on w.Write

sguiheux:
1. Description
1. Related issues
1. About tests
1. Mentions

@ovh/cds

### DIFF
--- a/engine/api/events.go
+++ b/engine/api/events.go
@@ -296,9 +296,15 @@ func (b *eventsBroker) ServeHTTP() Handler {
 
 		go func() {
 			for msg := range messageChan.Queue {
-				w.Write([]byte("data: "))
-				w.Write([]byte(msg))
-				w.Write([]byte("\n\n"))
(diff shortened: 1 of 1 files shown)


engine/api/events.go
```diff
@@ -296,9 +296,15 @@ func (b *eventsBroker) ServeHTTP() Handler {
 
 		go func() {
 			for msg := range messageChan.Queue {
-				w.Write([]byte("data: "))
-				w.Write([]byte(msg))
-				w.Write([]byte("\n\n"))
+				if _, err := w.Write([]byte("data: ")); err != nil {
```
yesnault: one write?

engine/api/events.go
```diff
@@ -13,6 +13,7 @@ import (
 
 	"github.com/go-gorp/gorp"
 
+	"bytes"
```
golangci: File is not goimports-ed

PR #2823, opened 2018-06-05
feat (api): add some debug logs about async

yesnault:

@ovh/cds


yesnault:
this will be useful to understand if error async is made by hatchery or worker (and which worker)

PR #2842, opened 2018-06-07
fix (api, ui): relaunch button

yesnault:
take care of pipeline parameters & payload (relaunch workflow)
take care of pipeline paremters (re-play pipeline)
delete unused dropdown from breadcrumb

### DIFF
--- a/ui/src/app/shared/workflow/node/run/node.run.param.component.ts
+++ b/ui/src/app/shared/workflow/node/run/node.run.param.component.ts
@@ -132,7 +132,7 @@ export class WorkflowNodeRunParamComponent {
             this.codeMirrorConfig = Object.assign({}, this.codeMirrorConfig, {readOnly: false});
         }
 
-        if (isRoot === true && this.workflowRun) {
+        if (isRoot && this.workflowRun) {
             let rootNodeRun = this.workflowRun.nodes[this.workflowRun.workflow.root.id][0];
             if (rootNodeRun && rootNodeRun.hook_event) {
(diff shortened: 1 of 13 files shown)


yesnault:
@fsamin Right
We need "HookEvent", "Manual" and only "PipelineParameters"
buildParameters is now with DisableDetailledNodeRun=false
It's not only for the root pipeline, this fix include relaunch of child pipelines (for pipeline_parameters)

ui/src/app/shared/workflow/node/run/node.run.param.component.ts
```diff
@@ -122,13 +122,27 @@ export class WorkflowNodeRunParamComponent {
...
         if (this.workflowRun && this.workflowRun.workflow.root_id !== this.nodeToRun.id) {
             this.readOnly = true;
             this.codeMirrorConfig = Object.assign({}, this.codeMirrorConfig, {readOnly: true});
         } else {
             this.readOnly = false;
+            isRoot = true;
             this.codeMirrorConfig = Object.assign({}, this.codeMirrorConfig, {readOnly: false});
         }
+
+        if (isRoot === true && this.workflowRun) {
```
sguiheux: remove  ->  === true

fsamin:
@yesnault 
and what about calling `getWorkflowNodeRunHandler` to get the detail on the node run when the model is opening ? as we do when we doubleclick on it.

So we could lighten `getWorkflowRunHandler` event more

sguiheux:
MIssing readonly mode on list parameter if we relaunch a noderun ?

PR #2856, opened 2018-06-10
feat (ui): add filter on artifact list

yesnault:
1. About tests
manual done


ui/src/app/views/workflow/run/node/artifact/artifact.list.component.ts
```diff
@@ -14,14 +14,18 @@ export class WorkflowRunArtifactListComponent extends Table {
...
         super();
         this.zone = new NgZone({enableLongStackTrace: false});
     }
 
     getData(): any[] {
-        return this.artifacts;
+        if (!this.filter) {
+            return this.artifacts;
+        }
+        return this.artifacts.filter(v => (v.name.indexOf(this.filter) !== -1 || v.sha512sum.indexOf(this.filter) !== -1));
```
bnjjj: I think in case of sha512 research it's better to use `===` instead of `indexOf`

PR #2860, opened 2018-06-11
fix(hatchery/swarm): don't try to remove docker bridge

### LINKED ISSUE #2733: Log in hatchery swarm on pre-defined network
Log in hatchery swarm
````
killAndRemove> unable to kill and remove network f0b4c78490bd err:Error response from daemon: bridge is a pre-defined network and cannot be removed
```

And there is no service link on pre-requisite on the job

fsamin:
close #2733

1. Description
1. Related issues
1. About tests
1. Mentions

@ovh/cds

### DIFF
--- a/engine/hatchery/swarm/swarm_util_kill.go
+++ b/engine/hatchery/swarm/swarm_util_kill.go
@@ -10,6 +10,11 @@ import (
 	"github.com/ovh/cds/sdk/log"
 )
 
+const (
+	bridge  = "bridge"
+	docker0 = "docker0"
+)
+
 func (h *HatcherySwarm) killAndRemove(ID string) error {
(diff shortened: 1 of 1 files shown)


engine/hatchery/swarm/swarm_util_kill.go
```diff
@@ -39,6 +39,11 @@ func (h *HatcherySwarm) killAndRemove(ID string) error {
 			continue
 		}
 
+		//If it's the default docker bridge... skip
+		if network.Driver != "bridge" || network.Name == "docker0" || network.Name == "bridge" {
```
golangci: string `bridge` has 4 occurrences, make it a constant

PR #2863, opened 2018-06-11
feat (hatchery/swarm): log on err create network

yesnault:


engine/hatchery/swarm/swarm.go
```diff
@@ -87,7 +87,9 @@ func (h *HatcherySwarm) SpawnWorker(spawnArgs hatchery.SpawnArguments) (string,
 				if network == "" {
 					network = name + "-net"
 					networkAlias = "worker"
-					h.createNetwork(network)
+					if err := h.createNetwork(network); err != nil {
```
sguiheux: Do we have to continue if network create failed ?

PR #2885, opened 2018-06-13
 fix(api): workflow as code and beyond

fsamin:
you can keep the commits and merge instead of squash

### DIFF
--- a/engine/api/application_import_test.go
+++ b/engine/api/application_import_test.go
@@ -599,19 +599,11 @@ func Test_postApplicationImportHandler_ExistingAppWithDeploymentStrategy(t *test
 	proj := assets.InsertTestProject(t, db, api.Cache, sdk.RandomString(10), sdk.RandomString(10), u)
 	test.NotNil(t, proj)
 
+	pfname := sdk.RandomString(10)
 	pf := sdk.PlatformModel{
-		Name:       "test-deploy-2",
+		Name:       pfname,
 		Deployment: true,
-	}
-	test.NoError(t, platform.InsertModel(db, &pf))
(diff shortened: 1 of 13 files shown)


engine/api/workflow/hook.go
```diff
@@ -248,3 +249,54 @@ func mergeAndDiffHook(oldHooks map[string]sdk.WorkflowNodeHook, newHooks map[str
...
 	return
 }
+
+// DefaultPayload returns the default payload for the workflow root
+func DefaultPayload(db gorp.SqlExecutor, store cache.Store, p *sdk.Project, u *sdk.User, wf *sdk.Workflow) (interface{}, error) {
+	var defaultPayload interface{}
+	if wf.Root.Context == nil || wf.Root.Context.Application == nil || wf.Root.Context.Application.ID == 0 {
+		app, errLa := application.LoadByID(db, store, wf.Root.Context.ApplicationID, u)
+		if errLa != nil {
+			return wf.Root.Context.DefaultPayload, sdk.WrapError(errLa, "getDefaultPayload> unable to load application by id %d", wf.Root.Context.ApplicationID)
```
bnjjj: Change log s/getDefaultPayload/DefaultPayload everywhere in this function

engine/api/application_import_test.go
```diff
@@ -622,6 +614,15 @@ func Test_postApplicationImportHandler_ExistingAppWithDeploymentStrategy(t *test
 			},
 		},
 	}
+	test.NoError(t, platform.InsertModel(db, &pf))
+	defer platform.DeleteModel(db, pf.ID)
```
golangci: Error return value of `platform.DeleteModel` is not checked

PR #2895, opened 2018-06-15
fix(ui): don't display 'no parameter' if there isn't any parameter

fsamin:
1. Description
1. Related issues
1. About tests
1. Mentions

@ovh/cds

### DIFF
--- a/ui/src/app/shared/workflow/node/run/node.run.param.html
+++ b/ui/src/app/shared/workflow/node/run/node.run.param.html
@@ -17,7 +17,7 @@
     </div>
     <div class="content">
         <div class="ui form">
-            <ng-container *ngIf="_nodeToRun">
+            <ng-container *ngIf="_nodeToRun && _nodeToRun.context.default_pipeline_parameters.length > 0 ">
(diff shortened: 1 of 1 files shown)


ui/src/app/shared/workflow/node/run/node.run.param.html
```diff
@@ -17,7 +17,7 @@
     </div>
     <div class="content">
         <div class="ui form">
-            <ng-container *ngIf="_nodeToRun">
+            <ng-container *ngIf="_nodeToRun && _nodeToRun.context.default_pipeline_parameters.length > 0 ">
```
bnjjj: `_nodeToRun && _nodeToRun.context && _nodeToRun.context.default_pipeline_parameters && _nodeToRun.context.default_pipeline_parameters.length > 0`

fsamin: we don't do that the line below: 
`                <app-parameter-list [project]="project" [parameters]="_nodeToRun.context.default_pipeline_parameters"
`

PR #2899, opened 2018-06-15
fix(api): lint

fsamin:
1. Description
1. Related issues
1. About tests
1. Mentions

@ovh/cds


engine/api/test/test.go
```diff
@@ -85,7 +85,7 @@ func SetupPG(t log.Logger, bootstrapFunc ...Bootstrapf) (*gorp.DbMap, cache.Stor
 	if err != nil {
 		t.Fatalf("Unable to connect to redis: %v", err)
 	}
-	event.Cache = store
+	event.Initialize(event.KafkaConfig{}, store)
```
golangci: Error return value of `event.Initialize` is not checked

fsamin: but tests...

PR #2958, opened 2018-06-26
feat(api): add new event on project variable warning

sguiheux:
1. Description
1. Related issues
1. About tests
1. Mentions

@ovh/cds

### DIFF
--- a/engine/api/warning/warning_project_variable_test.go
+++ b/engine/api/warning/warning_project_variable_test.go
@@ -16,7 +16,6 @@ import (
 	"github.com/ovh/cds/engine/api/test/assets"
 	"github.com/ovh/cds/engine/api/workflow"
 	"github.com/ovh/cds/sdk"
-	"regexp"
 )
 
 func TestMissingProjectVariablePipelineJob(t *testing.T) {
@@ -574,7 +573,7 @@ func TestUnusedProjectVariableWarningOnApplicationEvent(t *testing.T) {
 		NewVariable: sdk.Variable{
 			Name:  "foo",
 			Type:  "string",
-			Value: fmt.Sprintf("Welcome all", v.Name),
+			Value: "Welcome all",
 		},
 	}
 	eUpdate := sdk.Event{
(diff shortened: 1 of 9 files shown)


engine/api/warning/warning_project_variable_test.go
```diff
@@ -502,6 +503,93 @@ func TestMissingProjectVariableEnv(t *testing.T) {
...
+	ePayloadUpdate := sdk.EventApplicationVariableUpdate{
+		OldVariable: sdk.Variable{
+			Name:  "foo",
+			Type:  "string",
+			Value: fmt.Sprintf("Welcome {{.cds.proj.%s}}", v.Name),
+		},
+		NewVariable: sdk.Variable{
+			Name:  "foo",
+			Type:  "string",
+			Value: fmt.Sprintf("Welcome all", v.Name),
```
golangci: Sprintf call has arguments but no formatting directives

engine/api/warning/warning_project_variable_test.go
```diff
@@ -16,6 +16,7 @@ import (
 	"github.com/ovh/cds/engine/api/test/assets"
 	"github.com/ovh/cds/engine/api/workflow"
 	"github.com/ovh/cds/sdk"
+	"regexp"
```
golangci: File is not goimports-ed

engine/api/warning/warning_project_variable_test.go
```diff
@@ -573,3 +661,12 @@ func TestUnusedProjectVariableWarning(t *testing.T) {
 	test.NoError(t, errAfterDelete)
 	assert.Equal(t, 0, len(warnsDelete))
 }
+
+func TestToto(t *testing.T) {
+	re, err := regexp.Compile("cds\\.proj\\.[a-zA-Z\\-_]+")
```
golangci: should use raw string (`...`) with regexp.Compile to avoid having to escape twice

engine/api/warning/warning_variable.go
```diff
@@ -7,9 +7,74 @@ import (
...
+	"regexp"
+	"strings"
+	"time"
 )
 
+var projVarRegexp *regexp.Regexp
+
+func Init() error {
+	var err error
+	projVarRegexp, err = regexp.Compile("cds\\.proj\\.[a-zA-Z0-9\\-_]+")
```
golangci: should use raw string (`...`) with regexp.Compile to avoid having to escape twice

PR #2964, opened 2018-06-26
feat(api): async workflow deletion

fsamin:
1. Description
1. Related issues
1. About tests
1. Mentions

@ovh/cds

### DIFF
--- a/engine/api/purge/purge.go
+++ b/engine/api/purge/purge.go
@@ -1,0 +1,115 @@
+package purge
+
+import (
+	"context"
+	"database/sql"
+	"time"
+
+	"github.com/go-gorp/gorp"
+
+	"github.com/ovh/cds/engine/api/cache"
+	"github.com/ovh/cds/engine/api/project"
+	"github.com/ovh/cds/engine/api/workflow"
+	"github.com/ovh/cds/sdk"
+	"github.com/ovh/cds/sdk/log"
+)
+
+//Initialize starts goroutines for workflows
+func Initialize(c context.Context, store cache.Store, DBFunc func() *gorp.DbMap) {
+	tickPurge := time.NewTicker(30 * time.Second)
+	defer tickPurge.Stop()
+
+	for {
+		select {
+		case <-c.Done():
+			if c.Err() != nil {
+				log.Error("Exiting purge: %v", c.Err())
+				return
+			}
+		case <-tickPurge.C:
+			log.Debug("purge> Deleting all workflow run marked to delete...")
+			if err := RunsHistory(DBFunc()); err != nil {
+				log.Warning("purge> Error : %v", err)
+			}
+
+			log.Debug("purge> Deleting all workflow marked to delete....")
+			if err := Workflows(DBFunc(), store); err != nil {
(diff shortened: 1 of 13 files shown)


engine/api/workflow_run.go
```diff
@@ -687,16 +686,11 @@ func (api *API) postWorkflowRunHandler() Handler {
...
-					runtime.Stack(buf, true)
-					log.Error("[PANIC] workflow.PurgeWorkflowRun> %s", string(buf))
-				}
-			}()
-			workflow.PurgeWorkflowRun(api.mustDB(), *wf)
-		}()
+		sdk.GoRoutine(
+			"workflow.PurgeWorkflowRun",
+			func() {
+				workflow.PurgeWorkflowRun(api.mustDB(), *wf)
```
golangci: Error return value of `workflow.PurgeWorkflowRun` is not checked

engine/api/workflow/dao.go
```diff
@@ -629,11 +634,21 @@ func Update(db gorp.SqlExecutor, store cache.Store, w *sdk.Workflow, oldWorkflow
 	return updateLastModified(db, store, w, u)
 }
 
+// MarkAsDelete marks a workflow to be deleted
+func MarkAsDelete(db gorp.SqlExecutor, w *sdk.Workflow) error {
+	if _, err := db.Exec("update workflow set to_delete = true where id = $1", w.ID); err != nil {
+		return sdk.WrapError(err, "Delete> Unable to mark as delete workflow id %d", w.ID)
```
bnjjj: s/Delete/MarkAsDelete

fsamin:
you're assuming, that the user is able to delete a generated workflow (with a very few number of runs) and generated a new workflow faster than the goroutine is running at the end of `deleteWorkflowHandler` ?

i don't think it's possible and i think it's not useful to cover this nearly impossible usecase, especially because the 'migration' will be dropped in a few weeks.

yesnault:
@fsamin ok. The migrate use delete and not markAsDelete 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

engine/api/purge/purge.go
```diff
@@ -0,0 +1,115 @@
...
+
+	for _, w := range wrkflws {
+		tx, err := db.Begin()
+		if err != nil {
+			return sdk.WrapError(err, "purge.Workflows> unable to start tx")
+		}
+
+		proj, has := projects[w.ProjectID]
+		if !has {
+			continue
```
sguiheux: _ = tx.Rollback()

engine/api/purge/purge.go
```diff
@@ -0,0 +1,115 @@
...
+
+		if err := workflow.Delete(tx, store, &proj, &w); err != nil {
+			log.Error("purge.Workflows> unable to delete workflow %d: %v", w.ID, err)
+			_ = tx.Rollback()
+			continue
+		}
+
+		if err := tx.Commit(); err != nil {
+			log.Error("purge.Workflows> unable to commit tx: %v", err)
+			continue
```
sguiheux: _ = tx.Rollback()

PR #2967, opened 2018-06-27
fix(api): default payload from workflow as code

fsamin:
1. Description
1. Related issues
1. About tests
1. Mentions

@ovh/cds


fsamin:
this is a test

PR #2984, opened 2018-07-02
fix(api): fix repowebhook and workflow as code

bnjjj:

1. Description
1. Related issues
1. About tests
1. Mentions

@ovh/cds


bnjjj:
@fsamin ready to merge, tested with workflow as code and default payload

PR #3008, opened 2018-07-05
fix: better error code for duplicate key

sguiheux:
1. Description
Fix error code on duplicate key
1. Related issues
#2233 
1. About tests
1. Mentions

@ovh/cds

### DIFF
--- a/engine/api/environment/dao_key.go
+++ b/engine/api/environment/dao_key.go
@@ -5,6 +5,7 @@ import (
 	"encoding/base64"
 
 	"github.com/go-gorp/gorp"
+	"github.com/lib/pq"
 
 	"github.com/ovh/cds/engine/api/secret"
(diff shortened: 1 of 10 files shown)


engine/api/environment/dao_key.go
```diff
@@ -22,6 +23,9 @@ func InsertKey(db gorp.SqlExecutor, key *sdk.EnvironmentKey) error {
 	dbEnvironmentKey.Private = string(s)
 
 	if err := db.Insert(&dbEnvironmentKey); err != nil {
+		if errPG, ok := err.(*pq.Error); ok && errPG.Code == "23505" {
```
golangci: string `23505` has 4 occurrences, make it a constant

fsamin: nice catch @golangci !

PR #3013, opened 2018-07-06
Feat description icon

### LINKED ISSUE #2796: feat : prj / app / wf -> description, weight, icon
weight -> for drag / drop order in list

bnjjj:
1. Description
1. Related issues
close #2796
1. About tests
1. Mentions

@ovh/cds


engine/api/pipeline/pipeline.go
```diff
@@ -448,14 +448,14 @@ func UpdatePipeline(db gorp.SqlExecutor, p *sdk.Pipeline) error {
 	}
 
 	//Update pipeline
-	query := `UPDATE pipeline SET name=$1, type=$2 WHERE id=$3`
-	_, err := db.Exec(query, p.Name, string(p.Type), p.ID)
+	query := `UPDATE pipeline SET name=$1, description = $2, type=$3 WHERE id=$4`
+	_, err := db.Exec(query, p.Name, p.Description, string(p.Type), p.ID)
```
golangci: unnecessary conversion

engine/api/pipeline/pipeline.go
```diff
@@ -470,7 +470,7 @@ func InsertPipeline(db gorp.SqlExecutor, store cache.Store, proj *sdk.Project, p
 		return sdk.WrapError(sdk.ErrInvalidProject, "InsertPipeline>")
 	}
 
-	if err := db.QueryRow(query, p.Name, p.ProjectID, string(p.Type)).Scan(&p.ID); err != nil {
+	if err := db.QueryRow(query, p.Name, p.Description, p.ProjectID, string(p.Type)).Scan(&p.ID); err != nil {
```
golangci: unnecessary conversion

engine/api/project/dao_dependencies.go
```diff
@@ -184,6 +184,15 @@ var (
 		return nil
 	}
 
+	loadIcon = func(db gorp.SqlExecutor, store cache.Store, proj *sdk.Project, u *sdk.User) error {
```
fsamin: shouldn't it be better to load the icon directly to avoid executing an additional query ?

bnjjj: I think no, we load everywhere project, and icon could be 100ko and we just have to fetch for UI

PR #3029, opened 2018-07-12
fix(hatchery/marathon): func WorkerStarted should remove prefixes

fsamin:
1. Description
1. Related issues
1. About tests
1. Mentions

@ovh/cds


engine/hatchery/marathon/marathon.go
```diff
@@ -463,6 +455,9 @@ func (h *HatcheryMarathon) WorkersStarted() []string {
 	res := make([]string, len(apps))
 	for i, s := range apps {
 		res[i] = strings.Replace(s, h.Config.MarathonIDPrefix, "", 1)
+		if strings.HasPrefix(res[i], "/") {
```
golangci: should replace this if statement with an unconditional strings.TrimPrefix

PR #3045, opened 2018-07-18
 feat(worker): Install and use pgp key in a step script

### LINKED ISSUE #3023: Install and use pgp key in a step script

bnjjj:
1. Description
1. Related issues
close #3023
1. About tests
IT
1. Mentions

@ovh/cds

### DIFF
--- a/engine/worker/cmd_key.go
+++ b/engine/worker/cmd_key.go
@@ -45,9 +45,9 @@ func cmdKeyInstall(w *currentWorker) *cobra.Command {
 		Aliases: []string{"i", "add"},
 		Short:   "worker key install <key-name>",
 		Long: `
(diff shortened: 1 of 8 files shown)


engine/worker/cmd_key.go
```diff
@@ -142,27 +150,86 @@ func (wk *currentWorker) keyInstallHandler(w http.ResponseWriter, r *http.Reques
...
+			writeJSON(w, errC, errC.Status)
+			return
 		}
-		log.Error("%v", errSetup)
-		writeJSON(w, errSetup, errSetup.Status)
-		return
-	}
 
-	writeJSON(w, keyResponse{PKey: wk.currentJob.pkey}, http.StatusOK)
+		cmd := exec.Command("gpg", "--import", tmpfile.Name())
```
fsamin: try to lookupPath to raise a proper error if there is no gpg
what about gpg2 on some debians ?

PR #3057, opened 2018-07-23
feat(api, ui): de-duplicate pipelines in workflow

fsamin:
1. Description
Removing pipeline struct for workflowNode to avoid duplicate pipeline instances. Now the pipelines are a map at the workflow level
1. Related issues
1. About tests
1. Mentions

@ovh/cds

### DIFF
--- a/ui/src/app/model/workflow.model.ts
+++ b/ui/src/app/model/workflow.model.ts
@@ -370,7 +370,6 @@ export class Workflow {
     }
 
     static getPipeline(workflow: Workflow, node: WorkflowNode): Pipeline {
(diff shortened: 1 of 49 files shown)


ui/src/app/model/workflow.model.ts
```diff
@@ -368,6 +369,11 @@ export class Workflow {
         return {found: false};
     }
 
+    static getPipeline(workflow: Workflow, node: WorkflowNode): Pipeline {
+        console.log(workflow);
```
bnjjj: to delete

PR #3062, opened 2018-07-23
Fixed hatchery swarm

karantin2020:
1. The problem was in setting worker container network. By default created containers had no network. That's why they couldn't connect to cds-api and cds-hatchery-swarm containers. Connecting through host network is not always possible or real. Updated docker/docker vendor folder to have support of new docker api
2. Fixed worker containers network settings in hatchery swarm. Updated docker/docker vendor folder
3. All tests complete
4. Changes in docker-compose file compare to changes in cds-hatchery-swarm. It is not necessary to merge changes in docker-compose.yml in master branch

@ovh/cds


engine/hatchery/swarm/swarm.go
```diff
@@ -73,6 +74,23 @@ func (h *HatcherySwarm) SpawnWorker(spawnArgs hatchery.SpawnArguments) (string,
...
 	var network, networkAlias string
 	services := []string{}
+	if network == "" {
+		hn, err := os.Hostname()
+		if err != nil {
+			return "", err
+		}
+		if nl, err := getContainerNetworks(h.dockerClient, hn); err != nil {
+			return "", err
+		} else {
```
golangci: if block ends with a return statement, so drop this else and outdent its block (move short variable declaration to its own line if necessary)

engine/hatchery/swarm/swarm.go
```diff
@@ -601,3 +620,16 @@ func (h *HatcherySwarm) NeedRegistration(m *sdk.Model) bool {
 	}
 	return false
 }
+
+func getContainerNetworks(client *docker.Client, containerId string) ([]string, error) {
```
golangci: func parameter containerId should be containerID

engine/hatchery/swarm/swarm.go
```diff
@@ -601,3 +620,16 @@ func (h *HatcherySwarm) NeedRegistration(m *sdk.Model) bool {
...
+	var cj types.ContainerJSON
+	var err error
+	if cj, err = client.ContainerInspect(context.Background(), containerId); err != nil {
+		return nil, sdk.WrapError(err, "Unable to fetch container info")
+	}
+	nets := make([]string, 0, 4)
+	for name, _ := range cj.NetworkSettings.Networks {
```
golangci: should omit 2nd value from range; this loop is equivalent to `for name := range ...`

yesnault:
Thanks a lot for your pull request - sure, there is an issue with the default values for environment variables on docker swarm.

Details:

As you update :
      CDS_HATCHERY_SWARM_COMMONCONFIGURATION_API_HTTP_URL: http://${HOSTNAME}:8081	
to :
      CDS_HATCHERY_SWARM_COMMONCONFIGURATION_API_HTTP_URL: http://cds-api:8081
Workers need to have a network to reach cds-api too.

We will remove "cds-api" network from hatchery-swarm, it's confusing to keep it.  Fixed with https://github.com/ovh/cds/pull/3300/files

But, there is an issue on docker-compose. With the variable CDS_HATCHERY_SWARM_MAXCONTAINERS = 4 -> the hatchery can't spawn worker (as max containers is reached with cds-ui, cds-cache, cds-api, cds-hatchery-swarm, cds-db...). Fixed with https://github.com/ovh/cds/pull/3300 too

But, it remains your real usecase:
_"Connecting through host network is not always possible or real."_ 
For now, we don't want to add docker network on worker unless it's really necessary in a production case.
Can you explains more your use case? In our side, in production, we don't use docker-compose with a Swarm Hatchery, using cds-api network. In on all our CDS Installations, the workers always communicate with CDS API with a LB, an IP, and never with a docker network.

I close this pull request. Feel free to add comment, we will reopen an issue / PR if needed (adding cds-api network, creating network on worker as you suggested in your pull request).

PR #3069, opened 2018-07-24
fix(worker): add error line number #3068

### LINKED ISSUE #3068: feature request: give the position of a bad interpolation in script (syntax error)
When a script has a bad template format, the error message does not give the line:
`End of step /Script-1 [Fail] with reason: Unable to interpolate action parameters`

It would be nice to have more details especially when the script is long.

bnjjj:

1. Description
1. Related issues
close #3068 
1. About tests
1. Mentions
cc @pgaxatte 

@ovh/cds

### DIFF
--- a/engine/worker/run.go
+++ b/engine/worker/run.go
@@ -155,7 +155,7 @@ func (w *currentWorker) runJob(ctx context.Context, a *sdk.Action, buildID int64
 		return sdk.Result{
 			Status:  sdk.StatusFail.String(),
 			BuildID: buildID,
-			Reason:  "Unable to interpolate action parameters",
+			Reason:  err.Error(),
 		}
 	}
 	// Set the params


engine/worker/run.go
```diff
@@ -155,7 +155,7 @@ func (w *currentWorker) runJob(ctx context.Context, a *sdk.Action, buildID int64
 		return sdk.Result{
 			Status:  sdk.StatusFail.String(),
 			BuildID: buildID,
-			Reason:  "Unable to interpolate action parameters",
+			Reason:  err.Error(),
```
fsamin: i suggest: `"Unable to interpolate action parameters (%v)", err`

bnjjj: no, because in the error of interpolate there is already `Unable to interpolate action parameters` so in fact the final sentence is for example : `Unable to interpolate action parameters: Invalid template format: template: input:17`

PR #3131, opened 2018-08-02
feat (worker): add retry on upload swift if error

yesnault:
@ovh/cds

### DIFF
--- a/sdk/artifact.go
+++ b/sdk/artifact.go
@@ -354,9 +354,12 @@ func uploadArtifactWithTempURL(project, pipeline, application, env, tag string, 
 		}
 		resp.Body.Close()
 
-		if resp.StatusCode >= 300 {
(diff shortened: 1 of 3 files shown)


sdk/artifact.go
```diff
@@ -333,35 +333,47 @@ func uploadArtifactWithTempURL(project, pipeline, application, env, tag string,
...
+		if err != nil {
+			globalErrSwift = fmt.Errorf("Unable to upload artifact - try %d (Read Body) err: %v", i, err)
+			continue
+		}
+		resp.Body.Close()
 
-	body, err = ioutil.ReadAll(resp.Body)
-	if err != nil {
-		return true, 0, err
+		if resp.StatusCode >= 300 {
```
fsamin: only 500 error, >= 400 && < 500 will always fail i guess

PR #3145, opened 2018-08-06
feat: handle * on plugin venom for pathfile andexcluded pathfile

marcaudefroy:
Signed-off-by: Marc Audefroy <marc.audefroy@corp.ovh.com>

1. Description

Currently, plugin don't handle * into the path and excluded arguments. This PR add this behaviour.

1. Related issues
1. About tests
1. Mentions

@ovh/cds

### DIFF
--- a/contrib/plugins/plugin-venom/venom.go
+++ b/contrib/plugins/plugin-venom/venom.go
@@ -3,6 +3,7 @@ package main
 import (
 	"encoding/json"
 	"io/ioutil"
+	"os"
 	"path/filepath"
 	"strconv"
 	"strings"
@@ -188,8 +189,24 @@ func (s VenomPlugin) Run(a plugin.IJob) plugin.Result {
 		filepath = strings.Split(filepath[0], " ")
 	}
 
+	for _, fp := range filepath {
+		filepath, err = walkGlobFile(fp)
+		if err != nil {
+			plugin.SendLog(a, "VENOM - Error on walk files: %v\n", err) // nolint
+			return plugin.Fail
+		}
+	}
+
 	if len(filepathExcluded) == 1 {
 		filepathExcluded = strings.Split(filepathExcluded[0], " ")
+	}
+
+	for _, fp := range filepathExcluded {
+		filepathExcluded, err = walkGlobFile(fp)
+		if err != nil {
+			plugin.SendLog(a, "VENOM - Error on walk excluded files: %v\n", err) // nolint
+			return plugin.Fail
+		}
 	}
 
 	tests, err := v.Process(filepath, filepathExcluded)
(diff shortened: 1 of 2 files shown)


contrib/plugins/plugin-venom/venom.go
```diff
@@ -188,10 +189,26 @@ func (s VenomPlugin) Run(a plugin.IJob) plugin.Result {
 		filepath = strings.Split(filepath[0], " ")
 	}
 
+	for _, fp := range filepath {
+		filepath, err = walkGlobFile(fp)
+		if err != nil {
+			plugin.SendLog(a, "VENOM - Error on walk files: %v\n", err)
```
golangci: Error return value of `plugin.SendLog` is not checked

contrib/plugins/plugin-venom/venom.go
```diff
@@ -188,10 +189,26 @@ func (s VenomPlugin) Run(a plugin.IJob) plugin.Result {
...
+	}
+
 	if len(filepathExcluded) == 1 {
 		filepathExcluded = strings.Split(filepathExcluded[0], " ")
 	}
 
+	for _, fp := range filepathExcluded {
+		filepathExcluded, err = walkGlobFile(fp)
+		if err != nil {
+			plugin.SendLog(a, "VENOM - Error on walk excluded files: %v\n", err)
```
golangci: Error return value of `plugin.SendLog` is not checked

contrib/plugins/plugin-venom/venom.go
```diff
@@ -211,3 +228,24 @@ func (s VenomPlugin) Run(a plugin.IJob) plugin.Result {
 func main() {
 	plugin.Main(&VenomPlugin{})
 }
+
+func walkGlobFile(path string) ([]string, error) {
+	filenames, _ := filepath.Glob(path)
```
yesnault: err should be checked

marcaudefroy: done

contrib/plugins/plugin-venom/venom.go
```diff
@@ -211,3 +228,27 @@ func (s VenomPlugin) Run(a plugin.IJob) plugin.Result {
 func main() {
 	plugin.Main(&VenomPlugin{})
 }
+
+func walkGlobFile(path string) ([]string, error) {
+	filenames, err := filepath.Glob(path)
+	if err != nil {
+		return []string{}, err
```
bnjjj: why not `return nil, err` ?

contrib/plugins/plugin-venom/venom.go
```diff
@@ -211,3 +228,27 @@ func (s VenomPlugin) Run(a plugin.IJob) plugin.Result {
...
+				if err != nil {
+					return err
+				}
+				if !info.IsDir() {
+					result = append(result, path)
+				}
+				return nil
+			})
+		if err != nil {
+			return []string{}, err
```
bnjjj: why not `return nil, err` ?

PR #3149, opened 2018-08-06
fix(api): fix stop workflow run for purge

bnjjj:

1. Description
1. Related issues
1. About tests
1. Mentions

@ovh/cds

### DIFF
--- a/engine/api/workflow/dao_node_run.go
+++ b/engine/api/workflow/dao_node_run.go
@@ -658,8 +658,8 @@ func updateNodeRunCommits(db gorp.SqlExecutor, id int64, commits []sdk.VCSCommit
 	return nil
 }
 
-// UpdateNodeRunStatusAndStage update just noderun status and stage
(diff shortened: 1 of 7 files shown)


engine/api/workflow/dao_node_run.go
```diff
@@ -658,7 +658,8 @@ func updateNodeRunCommits(db gorp.SqlExecutor, id int64, commits []sdk.VCSCommit
 	return nil
 }
 
-func updateNodeRunStatusAndStage(db gorp.SqlExecutor, nodeRun *sdk.WorkflowNodeRun) error {
+// UpdateNodeRunStatusAndStage update just noderun status and stage
+func UpdateNodeRunStatusAndStage(db gorp.SqlExecutor, nodeRun *sdk.WorkflowNodeRun) error {
```
fsamin: I think `updateNodeRunStatusAndStage` should not be public, it's a very low-level function, only called by other private functions.

Maybe the goal achieved in the purge package querying and modifying workflowNoRun is also a bit low-level and should be in the workflow package.

PR #3159, opened 2018-08-07
 feat(service): enable metrics on all services

fsamin:
1. Description
1. Related issues
1. About tests
1. Mentions

@ovh/cds


engine/service/http.go
```diff
@@ -0,0 +1,104 @@
...
+	Options      map[string]string
+}
+
+// Accepted is a helper function used by asynchronous handlers
+func Accepted(w http.ResponseWriter) error {
+	const msg = "request accepted"
+	w.Header().Add("Content-Type", "text/plain")
+	w.Header().Add("Content-Length", fmt.Sprintf("%d", len(msg)))
+	w.WriteHeader(http.StatusAccepted)
+	w.Write([]byte(msg))
```
golangci: Error return value of `w.Write` is not checked

engine/service/http.go
```diff
@@ -0,0 +1,104 @@
...
+	return nil
+}
+
+// Write is a helper function
+func Write(w http.ResponseWriter, btes []byte, status int, contentType string) error {
+	w.Header().Add("Content-Type", contentType)
+	w.Header().Add("Content-Length", fmt.Sprintf("%d", len(btes)))
+	WriteProcessTime(w)
+	w.WriteHeader(status)
+	w.Write(btes)
```
golangci: Error return value of `w.Write` is not checked

PR #3164, opened 2018-08-07
fix(hatchery/swarm): don't rm too young networks

yesnault:


yesnault:
cc @marcaudefroy

engine/hatchery/swarm/swarm_util_kill.go
```diff
@@ -126,6 +127,11 @@ func (h *HatcherySwarm) killAwolNetworks() error {
 				continue
 			}
 
+			// if network created less than 5 min, keep it alive for now
+			if time.Now().Sub(nets[i].Created) < 5*time.Minute {
```
golangci: should use time.Since instead of time.Now().Sub

marcaudefroy:
@yesnault  : seriously? 11pm?

PR #3173, opened 2018-08-09
fix(test): fix unit tests

fsamin:
1. Description
1. Related issues
1. About tests
1. Mentions

@ovh/cds


engine/api/application/dao_test.go
```diff
@@ -17,6 +18,7 @@ import (
 
 func TestLoadByNameAsAdmin(t *testing.T) {
 	db, cache := test.SetupPG(t, bootstrap.InitiliazeDB)
+	event.Initialize(event.KafkaConfig{}, cache)
```
golangci: Error return value of `event.Initialize` is not checked

engine/api/migrate/migrate_workflow_test.go
```diff
@@ -18,6 +19,8 @@ import (
 
 func Test_ToWorkflow(t *testing.T) {
 	db, cache := test.SetupPG(t)
+	event.Initialize(event.KafkaConfig{}, cache)
```
golangci: Error return value of `event.Initialize` is not checked

engine/api/pipeline/pipeline_importer_test.go
```diff
@@ -74,6 +75,8 @@ func testImportUpdate(t *testing.T, db gorp.SqlExecutor, store cache.Store, tt t
 
 func TestImportUpdate(t *testing.T) {
 	db, cache := test.SetupPG(t, bootstrap.InitiliazeDB)
+	event.Initialize(event.KafkaConfig{}, cache)
```
golangci: Error return value of `event.Initialize` is not checked

PR #3300, opened 2018-09-08
chore(docker-compose): maxcontainers is too small

yesnault:
CDS_HATCHERY_SWARM_MAXCONTAINERS = 4 -> the hatchery can't spawn workers
 (as max containers is reached with cds-ui, cds-cache, cds-api, cds-hatchery-swarm, cds-db...).


richardlt:
Can we filter containers in CanSpawn func to count only those created by the hatchery ?

PR #3302, opened 2018-09-09
fix(api): Return no err if app is not linked to a repo

yesnault:


fsamin:
btw `fix(engine)` should be `fix(api)`

PR #3304, opened 2018-09-09
fix(engine): Clean code and fix heartbeat, reset count only if no error

richardlt:
1. Description
1. Related issues
1. About tests
1. Mentions

@ovh/cds


engine/hatchery/local/local.go
```diff
@@ -382,39 +387,34 @@ func (h *HatcheryLocal) killAwolWorkers() error {
 	h.Lock()
 	defer h.Unlock()
 
-	apiworkers, err := h.CDSClient().WorkerList()
+	apiWorkers, err := h.CDSClient().WorkerList()
 	if err != nil {
 		return err
 	}
 
+	mAPIWorkers := map[string]sdk.Worker{}
```
bnjjj: Add length

PR #3317, opened 2018-09-12
fix(sdk): name generator - add random

yesnault:
There is too many collisions with namegenerator.
This explains the non-detecting worker timeout and the running jobs
since many days in workflow node run jobs.

### DIFF
--- a/sdk/namesgenerator/namesgenerator_test.go
+++ b/sdk/namesgenerator/namesgenerator_test.go
@@ -1,0 +1,28 @@
+package namesgenerator
+
+import (
+	"fmt"
+	"strings"
+	"testing"
+)
+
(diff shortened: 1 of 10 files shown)


sdk/namesgenerator/namesgenerator_test.go
```diff
@@ -0,0 +1,28 @@
...
+	"strings"
+	"testing"
+)
+
+func TestNameFormat(t *testing.T) {
+	name := GetRandomNameCDS(0)
+	if !strings.Contains(name, "_") {
+		t.Fatalf("Generated name does not contain an underscore")
+	}
+	fmt.Println("####### name:", name)
```
fsamin: t.log

PR #3319, opened 2018-09-13
feat(api,ui): add state filter on worker model

bnjjj:

1. Description
1. Related issues
1. About tests
1. Mentions

@ovh/cds

### DIFF
--- a/engine/api/worker/model.go
+++ b/engine/api/worker/model.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"sort"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/go-gorp/gorp"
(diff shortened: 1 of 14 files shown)


engine/api/worker/model.go
```diff
@@ -187,6 +217,10 @@ func LoadWorkerModelsByUser(db gorp.SqlExecutor, store cache.Store, user *sdk.Us
 					select %s from worker_model
 					JOIN "group" on worker_model.group_id = "group".id
 					where group_id = $2`, modelColumns, modelColumns)
+		if len(additionalFilters) > 0 {
+			query = fmt.Sprintf("%s AND %s", query, strings.Join(additionalFilters, " AND "))
```
fsamin: concat strings

engine/api/worker/model.go
```diff
@@ -54,6 +55,14 @@ type dbResultWMS struct {
 	GroupName string `db:"groupname"`
 }
 
+// LoadOptions represent load options to load worker model
+type LoadOptions struct {
```
richardlt: Why not a state type with four constants ?

PR #3427, opened 2018-10-09
feat(hatchery): allow force disable building workers (inconsistency)

yesnault:


yesnault:
yes, seen

PR #3518, opened 2018-10-29
fix(api): import pipeline error details

### LINKED ISSUE #3209: import job, with an action containing a deleted attribute
example:

p191-prod-api         | 2018-08-16 16:56:34 | warn | POST    | 500  | /project/FOOO/import/pipeline?format=yaml&forceUpdate=true 	 importPipelineHandler> Unable import pipeline: Job build validation Failure: Invalid parameter details on step #6 plugin-venom

--> message "Unable import pipeline: Job build validation Failure: Invalid parameter details on step #6 plugin-venom" should be returned to user

fsamin:
1. Description
It was only about error message management
1. Related issues
closes #3209
1. About tests
1. Mentions

@ovh/cds

### DIFF
--- a/sdk/error.go
+++ b/sdk/error.go
@@ -582,25 +582,32 @@ func callers() *stack {
 	return &st
 }
 
-// NewError just set an error with a root cause.
+// NewError returns a merge of given err with new http error.
 func NewError(httpError Error, err error) error {
-	// if the given error is a error with stack, replace the http error
+	// if the given error is nil do nothing
 	if err == nil {
 		return nil
 	}
 
+	// if it's already an error with stack, override the http error and set from value with err cause
 	if e, ok := err.(errorWithStack); ok {
 		httpError.from = Cause(e).Error()
 		e.httpError = httpError
 		return e
 	}
 
+	// if it's a library error create a new error with stack
 	httpError.from = err.Error()
 	return errorWithStack{
 		root:      errors.WithStack(err),
 		stack:     callers(),
 		httpError: httpError,
 	}
+}
+
+// NewErrorFrom returns the given http error with from details.
+func NewErrorFrom(httpError Error, from string, args ...interface{}) error {
+	return NewError(httpError, fmt.Errorf(from, args...))
 }
 
 // WrapError returns an error with stack and message.
--- a/sdk/error.go
+++ b/sdk/error.go
@@ -588,6 +588,15 @@ func NewError(httpError Error, err error) error {
 		if e, ok := err.(errorWithStack); ok {
 			e.httpError = httpError
 			return e
+		} else {
+			return errorWithStack{
+				root:  err,
+				stack: callers(),
+				httpError: Error{
+					Status:  httpError.Status,
+					Message: err.Error(),
+				},
+			}
 		}
 	}
 
--- a/sdk/error.go
+++ b/sdk/error.go
@@ -466,6 +466,7 @@ type Error struct {
 	Message    string `json:"message"`
 	UUID       string `json:"uuid,omitempty"`
 	StackTrace string `json:"stack_trace,omitempty"`
+	from       string
 }
 
 func (e Error) Error() string {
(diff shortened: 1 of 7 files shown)


ovh-cds:
CDS Report it#6777.0 ✘  
* w-it
  * w-it ✘

ovh-cds:
CDS Report ut-engine#6777.0 ✘  
* Stage 1
  * Engine Test with DB ✘  
Unit Tests Report
* github.com/ovh/cds/cli


* github.com/ovh/cds/engine/api

































































































































































































* github.com/ovh/cds/engine/api/application








* github.com/ovh/cds/engine/api/environment




* github.com/ovh/cds/engine/api/keys



* github.com/ovh/cds/engine/api/migrate


* github.com/ovh/cds/engine/api/pipeline






* github.com/ovh/cds/engine/api/plugin


* github.com/ovh/cds/engine/api/project









* github.com/ovh/cds/engine/api/secret






* github.com/ovh/cds/engine/api/user








* github.com/ovh/cds/engine/api/warning















* github.com/ovh/cds/engine/api/worker











* github.com/ovh/cds/engine/api/workflow














































* github.com/ovh/cds/engine/hatchery/swarm



















* github.com/ovh/cds/engine/hooks






* github.com/ovh/cds/engine/repositories


* github.com/ovh/cds/engine/vcs











* github.com/ovh/cds/engine/vcs/bitbucket


















* github.com/ovh/cds/engine/vcs/github













* github.com/ovh/cds/engine/vcs/gitlab









* github.com/ovh/cds/engine/worker






* github.com/ovh/cds/sdk








  * TestNewError ✘









* github.com/ovh/cds/sdk/cdsclient



* github.com/ovh/cds/sdk/exportentities





























* github.com/ovh/cds/sdk/interpolate
































* github.com/ovh/cds/sdk/luascript





* github.com/ovh/cds/sdk/namesgenerator



* github.com/ovh/cds/sdk/vcs/git

sdk/error.go
```diff
@@ -544,6 +544,15 @@ func NewError(httpError Error, err error) error {
 		if e, ok := err.(errorWithStack); ok {
 			e.httpError = httpError
 			return e
+		} else {
```
richardlt: This will complict with the error PR and maybe you want to do same thing

fsamin: i've just rebased my branch on master. No conflict on this.
I guess the master didn't match my needs because i have had to update unit tests to specifically check this.

here is the result on this issue (the job name is "New Job" and the action is "myAction")
screen shot 2018-10-31 at 08 50 58

richardlt: If a understand the needs here, we want to override the origin error message for given sdk.Error. 
There can be two types of error given in NewError second's parameter, a errorWithStack and other error (sdk.Error, or lib error). What's important is to always store the second parameter as the root to preserve a valid stack trace. On a errorWithStack we can just override the http error then change the message on in with the Cause of the error.
For other types set the root and print the error as the message for http error.

Something like this ?
`
// NewError just set an error with a root cause.
func NewError(httpError Error, err error) error {
	// if the given error is a error with stack, replace the http error
	if err == nil {
    return nil
  }
	
  if e, ok := err.(errorWithStack); ok {
			e.httpError = httpError
      e.httpError.Message = Cause(err)
			return e
	}

  e:= errorWithStack{
	 	 root:      errors.WithStack(err),
		 stack:     callers(),
		 httpError: httpError,
	}
  e.httpError.Message = Cause(err)
  return err
}
`

In this case all errors from NewError will contains the root error text as message and that's what we will see on the client for or cdsctl. In this case, printing the (caused by: ...) in ExtractHTTPError will not be needed anymore.

What do you think ?

fsamin: a lot of code is using the sdk.NewError function to override the message of an sdk.Error. All the import and as-code stuff relies on this.
i don't understand all the details on the new error management, but the purpose of the related issue is to raise the detailed message to the end user.

richardlt: Ok so the master version is not good :/, because is doesn't override the message but add a caused by next the original error translated message. So I will commit on this PR to fix this and update the test case.

PR #3528, opened 2018-10-31
refacto(api): workflow notification code and remove old workflows notifications business code

fsamin:
1. Description
In order to implement notification in the workflow as code format, it would like to simplify the users notifications model. Which is still stuck to the old workflows.
This is the first step, removing all the old notifications code in the API, deprecate the handler (without removing the routes, nor impacting the UI)
1. Related issues
1. About tests
1. Mentions

@ovh/cds


fsamin:
yes that's why a made this PR before and  i won't touch this branch anymore.
Workflow as-code notification will be in another PR based on yours.

PR #3549, opened 2018-11-05
fix(contrib): remove the right socket after usage of plugin

bnjjj:

1. Description
1. Related issues
1. About tests
1. Mentions

@ovh/cds


engine/worker/builtin.go
```diff
@@ -114,7 +113,7 @@ func (w *currentWorker) runGRPCPlugin(ctx context.Context, a *sdk.Action, buildI
 
 		defer func() {
 			if strings.Contains(pluginSocket.Socket, ".sock") {
-				if err := os.RemoveAll(socketPath); err != nil {
```
richardlt: socketPath value never set ?

bnjjj: Nope

yesnault:
done with another pr

PR #3556, opened 2018-11-06
fix(ui): display all logs after a click when job is running

### LINKED ISSUE #3553: Logs always hidden when a pipeline is running
A new feature to hide logs by default has been added some time ago.
This feature works well if a pipeline is completed or stopped.
However when trying to expand a log on a running pipeline
logs will expand and get hidden straight away.

When the decision has been made to show logs, logs should remain
expanded.

bnjjj:

1. Description
1. Related issues
close #3553 
1. About tests
1. Mentions

@ovh/cds

### DIFF
--- a/ui/angular.json
+++ b/ui/angular.json
@@ -191,6 +191,9 @@
             "tsConfig": [
               "src/tsconfig.lint.json"
             ],
+            "files": [
+              "src/**/*.ts"
+            ],
             "exclude": [
               "src/**/*.spec.ts",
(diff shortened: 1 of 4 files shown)


richardlt:
What will happen is the case when I'm looking at logs from the beginning of the job and lines count reach the value that triggers logs folding ?

bnjjj:
@richardlt it will fold logs, you can just click on "display all logs" to see all the logs

PR #3571, opened 2018-11-08
Add some new features for debpacker

richardlt:
1. Description
Allow to copy folder recursively with custom out dir
Allow to set the working directory
Allow to use root user
Auto enable created service
Add unit test for packer


richardlt:
Multiple changes, allow to copy files, use root user (don't create a new one), set the working directory.

fsamin:
thanks @richardlt  , could you add `test`goal in the make file in order to run the tests and generate a xUnit repost ? We also add to update the CDS workflow to run it

PR #3576, opened 2018-11-09
fix(api): Load WorkerModel

fsamin:
1. Description
1. Related issues
1. About tests
1. Mentions

@ovh/cds

### DIFF
--- a/sdk/worker.go
+++ b/sdk/worker.go
@@ -116,8 +116,8 @@ type Model struct {
 	GroupID                int64               `json:"group_id" db:"group_id" cli:"-"`
 	Group                  Group               `json:"group" db:"-" cli:"-"`
 	NbSpawnErr             int64               `json:"nb_spawn_err" db:"nb_spawn_err" cli:"nb_spawn_err"`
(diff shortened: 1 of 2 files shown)


sdk/worker.go
```diff
@@ -116,8 +116,8 @@ type Model struct {
 	GroupID                int64               `json:"group_id" db:"group_id" cli:"-"`
 	Group                  Group               `json:"group" db:"-" cli:"-"`
 	NbSpawnErr             int64               `json:"nb_spawn_err" db:"nb_spawn_err" cli:"nb_spawn_err"`
-	LastSpawnErr           string              `json:"last_spawn_err" db:"last_spawn_err" cli:"-"`
-	LastSpawnErrLogs       *string             `json:"last_spawn_err_log" db:"last_spawn_err_log" cli:"-"`
+	LastSpawnErr           string              `json:"last_spawn_err" db:"-" cli:"-"`
```
yesnault: it's normal to remove db column here (no postSelect on it) ?

Shouldn't be: 
```
	LastSpawnErr           string              `json:"last_spawn_err" db:"last_spawn_err" cli:"-"`
```
?

fsamin: correct, it has been fixed in the postGet

Issue #3166, opened 2018-08-08
ASCII colors in job output are not displayed


wilfriedroset:
Looks like CDS doesn't display colored text when displaying jobs output.
My job looks like this

```
NC='\033[0m' # No Color
RED='\033[0;31m'
GREEN='\033[0;32m'
echo -e "${GREEN}OK${NC}"
```

The output display correctly OK but the green color aren't displayed.

maxatome:
Output is HTML at the time of writing. So you can write HTML `font` tags as a workaround, or pipe your output to a script that does that for you.

What would be useful, is to have the choice of the output format for each step between raw, ANSI or HTML for example.

bnjjj:
Ability to add option on step to tell what kind of format the logs are. In that case when in UI we display logs, we know the preferred format et so adapt our display

yesnault:
close about display ascii.

@maxatome #3643 create for your request

PR #3579, opened 2018-11-12
fix(ui): ansi in logs is now displayed (#3166)

bnjjj:

1. Description
1. Related issues
1. About tests
1. Mentions

@ovh/cds


richardlt:
Ok, ansi_up with default mode has escape_for_html active. So logs that contains html will be escaped before render.

PR #3586, opened 2018-11-13
fix(ui): fix undefined parameters with fork launch

bnjjj:

1. Description
1. Related issues
1. About tests
1. Mentions

@ovh/cds


ui/src/app/shared/workflow/node/run/node.run.param.html
```diff
@@ -20,7 +20,7 @@
             <ng-container *ngIf="_nodeToRun && parameters?.length > 0 ">
```
sguiheux: && _nodeToRun.type === WNode.Pipeline

PR #3589, opened 2018-11-13
fix(api): catch locked element and add retry in sdk if it's locked (close #3581)

### LINKED ISSUE #3581: fix(worker): add retry on postWorkflowJobTestsResultsHandler
```pprofLabel>recoverWrap>Handle>postWorkflowJobTestsResultsHandler>LoadAndLockNodeRunByID: internal server error (caused by: Cannot load node job: Unable to load workflow_node_run node=2604599: pq: could not obtain lock on row in relation "workflow_node_run")```

we should handle the pq error `pq: could not obtain lock on row in relation "workflow_node_run"`to handle retry on all client function retries on this kind of error at client http level

```go
func (c *client) QueueSendResult(ctx context.Context, id int64, res sdk.Result) error {
	path := fmt.Sprintf("/queue/workflows/%d/result", id)
	_, err := c.PostJSON(ctx, path, res, nil)
	return err
}
```

bnjjj:

1. Description
1. Related issues
1. About tests
1. Mentions

@ovh/cds

### DIFF
--- a/engine/worker/builtin_junit.go
+++ b/engine/worker/builtin_junit.go
@@ -6,9 +6,11 @@ import (
 	"encoding/xml"
 	"fmt"
 	"io/ioutil"
+	"net/http"
 	"net/url"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/ovh/cds/sdk"
 	"github.com/ovh/venom"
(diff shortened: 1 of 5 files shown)


engine/worker/builtin_junit.go
```diff
@@ -93,13 +95,21 @@ func runParseJunitTestResultAction(w *currentWorker) BuiltInAction {
...
+		var errPost error
+		for retry := 0; retry < 10; retry++ {
+			_, statusCode, errPost = sdk.Request("POST", uri, []byte(dataS))
+			if statusCode != http.StatusConflict {
+				break
+			}
+			time.Sleep(500 * time.Millisecond)
+		}
+		if errPost == nil && statusCode > 300 {
+			err = fmt.Errorf("HTTP %d", statusCode)
```
golangcibot: ineffectual assignment to `err`

PR #3642, opened 2018-11-23
fix(api): artifactDownload as code

### LINKED ISSUE #3637: fix(sdk): artifact download as code
i miss the pattern parameter :/
and there should be only `artifactDownload: <pattern>`
screen shot 2018-11-22 at 16 43 53

yesnault:
close #3637

### DIFF
--- a/sdk/exportentities/action.go
+++ b/sdk/exportentities/action.go
@@ -109,6 +109,14 @@ func newSteps(a sdk.Action) []Step {
 				if application != nil {
 					artifactDownloadArgs["application"] = application.Value
 				}
+				pattern := sdk.ParameterFind(&act.Parameters, "pattern")
+				if pattern != nil {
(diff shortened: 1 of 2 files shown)


sdk/exportentities/action.go
```diff
@@ -109,6 +109,14 @@ func newSteps(a sdk.Action) []Step {
 				if application != nil {
 					artifactDownloadArgs["application"] = application.Value
 				}
+				pattern := sdk.ParameterFind(&act.Parameters, "pattern")
+				if pattern != nil {
+					artifactDownloadArgs["pattern"] = pattern.Value
+				}
+				enabled := sdk.ParameterFind(&act.Parameters, "enabled")
+				if enabled != nil && enabled.Value == "false" {
```
richardlt: Why setting this only if false ?

yesnault: default value -> no attr on yml

PR #3644, opened 2018-11-24
feat(cdsctl): cdsctl admin service delete --name ...

yesnault:
need #3635 to add it


cli/cdsctl/admin_services.go
```diff
@@ -131,3 +132,20 @@ func adminServiceGetRun(v cli.Values) error {
 	fmt.Println(string(btes))
 	return nil
 }
+
+var adminServiceDeleteCmd = cli.Command{
+	Name:  "delete",
+	Short: "Delete a CDS service from registered service",
+	Flags: []cli.Flag{
```
fsamin: it should be a variadic arg instead of of a flag

yesnault: done

PR #3651, opened 2018-11-27
add .golangci.yml

jirfag:
It's needed to properly run analysis on https://golangci.com

Now analysis fails because of not installed `libsecret-1`. With this fix, it's analyzed [successfully](https://golangci.com/r/github.com/jirfag/cds).

@ovh/cds


yesnault:
@jirfag thanks a lot !

Issue #3714, opened 2018-12-10
fix(api): migrate previous worker model


bnjjj:
migrate previous worker model with {{.PipelineBuildJobID}} in udata

PR #3715, opened 2018-12-10
fix(sql): remove potential pb flags

yesnault:


yesnault:
close #3714

Issue #3702, opened 2018-12-06
no more pull requests comments about failing workflows


fsamin:
since about 3 weeks

fsamin:
bug is here, linked to the new entities: 
https://github.com/ovh/cds/blob/6c9476604c70ec2b15ac6222561d4f1d1c14514b/engine/api/workflow/workflow_run_event.go#L193

seen with @sguiheux 
poke @arnaudmorin

PR #3722, opened 2018-12-12
fix(api): fix send pullrequest report + do not load node from old model

sguiheux:
1. Description
1. Related issues
1. About tests
1. Mentions

@ovh/cds


fsamin:
related to #3702

PR #3775, opened 2018-12-26
fix(cdsctl): ask for context in shell only if current directory is a repository

richardlt:
1. Description
1. Related issues
1. About tests
1. Mentions

@ovh/cds

### DIFF
--- a/cli/cdsctl/shell.go
+++ b/cli/cdsctl/shell.go
@@ -94,8 +94,8 @@ func shellRun(v cli.Values) error {
 	home := "/"
 
 	// try to discover conf for existing .git repository
-	r, err := repo.New(".")
-	if err == nil {
+	r, errR := repo.New(".")
+	if errR == nil {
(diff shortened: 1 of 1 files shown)


cli/cdsctl/shell.go
```diff
@@ -85,21 +85,22 @@ func shellRun(v cli.Values) error {
...
-	// auto-discover current project with .git
-	if _, err := discoverConf([]cli.Arg{
-		{Name: _ProjectKey},
-		{Name: _ApplicationName, AllowEmpty: true},
-		{Name: _WorkflowName, AllowEmpty: true},
-	}); err == nil {
-		if r, err := repo.New("."); err == nil {
+
+	// try to discover conf for existing .git repository
+	r, err := repo.New(".")
```
bnjjj: should be named differently than the err inside if

richardlt: done

PR #3799, opened 2019-01-03
fix(api): keep spawnInfos on stop

### LINKED ISSUE #3699: fix(api): spawn infos disapear
Steps to reproduce : 
+ Have a job with a requirement model and a binary which is not in the model
+ Launch this job
+ You can see multiple spawn infos and several which indicate that our binary isn't good
+ Stop the job. Then the spawn infos disapear

yesnault:
close #3699

### DIFF
--- a/engine/api/workflow/execute_node_run.go
+++ b/engine/api/workflow/execute_node_run.go
@@ -692,13 +692,22 @@ func stopWorkflowNodePipeline(ctx context.Context, dbFunc func() *gorp.DbMap, st
 	}
 	wg.Wait()
 
+	tx, errTx := dbFunc().Begin()
+	if errTx != nil {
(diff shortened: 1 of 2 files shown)


engine/api/workflow/execute_node_run.go
```diff
@@ -693,7 +693,7 @@ func stopWorkflowNodePipeline(ctx context.Context, dbFunc func() *gorp.DbMap, st
 	wg.Wait()
 
 	// Update stages from node run
-	stopWorkflowNodeRunStages(nodeRun)
+	stopWorkflowNodeRunStages(dbFunc(), nodeRun)
```
fsamin: we should put a sql transaction in there

PR #3888, opened 2019-01-23
feat(ui) Job variable into real scrollable modal

svandecappelle:
1. Description
Set jobs variable modals size fit to screen & scroll its content
1. Related issues
1. About tests
1. Mentions

@ovh/cds


svandecappelle:
> Can we use that instead of overriding semantic style ? https://semantic-ui.com/modules/modal.html#scrolling-content

Cause the classname scrolling does nothing. Don't know why. Also the attribute className: {scrolling: 'scrolling' } doesn't works. 
Saw here: https://ng-semantic.herokuapp.com/#/elements/modal and https://semantic-ui.com/modules/modal.html#/settings

PR #4004, opened 2019-03-07
feat(ui): allow to init a new action from pipeline job

richardlt:
1. Description
screen

1. Related issues
1. About tests
1. Mentions

@ovh/cds


fsamin:
my opinion is that if I would like to create an action, i'll go to the action page, then eventually I'll choose to pickup something from a pipeline. This seems to me more natural rather than the creating an action from the pipeline

richardlt:
Up.

screen

PR #4014, opened 2019-03-11
fix(api): use new context for async part

sguiheux:
1. Description
1. Related issues
1. About tests
1. Mentions

@ovh/cds


engine/api/workflow_run.go
```diff
@@ -849,7 +849,7 @@ func (api *API) postWorkflowRunHandler() service.Handler {
 		}
 
 		// Purge workflow run
```
fsamin:
```suggestion
		// Workflow Run initialization
```

PR #4062, opened 2019-03-20
fix(api): check worker model requirements on job processing

### LINKED ISSUE #4002: fix:permissions on worker models
When selecting a worker model in the requirement of a pipeline, I should not be able to select the worker models not allowed on the project

For instance : 
- a user **U** in group **A**
- a worker model **WM** in the group **A**
- A project **P** on which Group **B** (only) has permission

If **U** creates a pipeline in **P**,
he can (in the pipeline he is editing) select **WM**
But when launching his pipeline, it will remains in idle state as **WM** is not allowed on **P** (no permission for **A**)

fsamin:
while computing the new job `addJobsToQueue`, the function `processNodeJobRunRequirements` now checks:
* the group of the worker model againt the `execGroups` computed on the node (close #4002)
* the binaries requirements against the worker model binary capabilitites
In case of error, the job is failed

Related issues:
#4002

About tests:
done manually

Mentions:
@ovh/cds

### DIFF
--- a/engine/api/workflow/process_requirements.go
+++ b/engine/api/workflow/process_requirements.go
@@ -70,6 +70,7 @@ func processNodeJobRunRequirements(db gorp.SqlExecutor, j sdk.Job, run *sdk.Work
 						for _, cap := range wm.RegisteredCapabilities {
 							if cap.Value == req.Value {
 								hasCapa = true
+                break
 							}
 						}
 						if !hasCapa {
--- a/engine/api/workflow/process_requirements.go
+++ b/engine/api/workflow/process_requirements.go
@@ -12,18 +12,14 @@ import (
 	"github.com/ovh/cds/sdk/log"
 )
 
-// getNodeJobRunRequirements returns requirements list interpolated, and true or false if at least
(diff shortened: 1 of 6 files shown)


ovh-cds:
CDS Report it#8876.1 ✘  
* w-it
  * w-it ✘

engine/api/workflow/process_requirements.go
```diff
@@ -36,29 +32,63 @@ func getNodeJobRunRequirements(db gorp.SqlExecutor, j sdk.Job, run *sdk.Workflow
...
+
+			// Check that the worker model has the binaries capabilitites
+			// only if the worker model doesn't need registration
+			if !wm.NeedRegistration && !wm.CheckRegistration {
+				for _, req := range requirements {
+					if req.Type == sdk.BinaryRequirement {
+						var hasCapa bool
+						for _, cap := range wm.RegisteredCapabilities {
+							if cap.Value == req.Value {
+								hasCapa = true
```
yesnault: if hasCapa == true -> could add a break to avoid range over all RegisteredCapabilities

Issue #3941, opened 2019-02-18
fix audit


fsamin:
* do not store audit for "as-code" entities (except workflow templates)
* limit to 30 audits
* delete old audits

fsamin:
cf #3954

PR #4071, opened 2019-03-22
fix(api): purge pipeline audits

fsamin:
1. Description
Keep the last 50 audits
1. Related issues
#3941
1. About tests
1. Mentions

@ovh/cds

### DIFF
--- a/engine/api/pipeline/audit.go
+++ b/engine/api/pipeline/audit.go
@@ -1,13 +1,35 @@
 package pipeline
 
 import (
+	"context"
 	"encoding/json"
 	"time"
 
 	"github.com/go-gorp/gorp"
 
 	"github.com/ovh/cds/sdk"
+	"github.com/ovh/cds/sdk/log"
 )
+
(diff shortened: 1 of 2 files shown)


engine/api/pipeline/audit.go
```diff
@@ -109,3 +131,38 @@ func (p *PipelineAudit) PostInsert(s gorp.SqlExecutor) error {
...
+				log.Error("purgeAudits> unable to delete audit %d: %v", id, err)
+			}
+		}
+	}
+
+	return nil
+}
+
+func deleteAudit(db gorp.SqlExecutor, id int64) error {
+	_, err := db.Exec(`delete from pipeline_audit where id = $1`, id)
```
richardlt: where id = ANY... we can remove all the audit for each pipeline in one time to reduce request count.

sguiheux: I think it's to avoid statement timeout if there is too much audits

fsamin: yep

PR #4072, opened 2019-03-22
fix(api): error inserting UpdateSpawnErrorWorkerModel with UTF8 0x00

### LINKED ISSUE #4055: WorkerModelSpawnError: internal server error
```
019-03-19 17:13:40 [WARN] pprofLabel>recoverWrap>Handle>spawnErrorWorkerModelHandler>UpdateSpawnErrorWorkerModel: internal server error (caused by: cannot update spawn error on worker model: pq: invalid byte sequence for encoding "UTF8": 0x00) error_uuid=f5b61728-4a61-11e9-9670-6a0002fa6430 method=PUT request_uri=/worker/model/error/1 stack_trace=pq: invalid byte sequence for encoding "UTF8": 0x00
```

fsamin:
1. Description
1. Related issues
close #4055
1. About tests
1. Mentions

@ovh/cds

### DIFF
--- a/engine/api/worker/model.go
+++ b/engine/api/worker/model.go
@@ -426,6 +426,7 @@ func updateAllToCheckRegistration(db gorp.SqlExecutor) error {
 
 // UpdateSpawnErrorWorkerModel updates worker model error registration
 func UpdateSpawnErrorWorkerModel(db gorp.SqlExecutor, modelID int64, spawnError sdk.SpawnErrorForm) error {
+  // some times when the docker container fails to start, the docker logs is not empty but only contains utf8 null char
 	if spawnError.Error == string([]byte{0x00}) {
 		spawnError.Error = ""
 	}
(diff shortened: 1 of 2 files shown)


engine/api/worker/model.go
```diff
@@ -426,6 +426,10 @@ func updateAllToCheckRegistration(db gorp.SqlExecutor) error {
 
 // UpdateSpawnErrorWorkerModel updates worker model error registration
 func UpdateSpawnErrorWorkerModel(db gorp.SqlExecutor, modelID int64, spawnError sdk.SpawnErrorForm) error {
+	if spawnError.Error == string([]byte{0x00}) {
```
richardlt: When is this happening ? Can you add a comment ?

fsamin: some times when the docker container fails to start, the docker logs is not empty but only contains utf8 null char

fsamin: comment added

PR #4078, opened 2019-03-22
fix(api): invalid payload error is not correctly raised

### LINKED ISSUE #4074: Cannot create a workflow outgoing hook without payload
When creating the workflow below, an error is thrown.

The error is:
Bad Request
Unable to update/delete the root node of your workflow

If the payload is set as '{}' the hook is created properly.

```
name: wf1
version: v1.0
workflow:
  Workflow:
    depends_on:
    - sleep-2s
    when:
    - success
    trigger: Workflow
    config:
      payload: ''
      target_hook: hook-id
      target_project: myproj
      target_workflow: wf2
  sleep-2s:
    when:
    - success
    pipeline: sleep-2s
```

Pipeline code:

```
version: v1.0
name: sleep-2s
jobs:
- job: sleep 2s
  steps:
  - script:
    - '#!/bin/bash'
    - ""
    - sleep 2
```

fsamin:
1. Description
1. Related issues
close #4074
1. About tests
1. Mentions
@ovh/cds

### DIFF
--- a/engine/api/workflow/dao.go
+++ b/engine/api/workflow/dao.go
@@ -949,7 +949,7 @@ func Update(ctx context.Context, db gorp.SqlExecutor, store cache.Store, w *sdk.
 	w.ResetIDs()
 
 	if err := insertNode(db, store, w, w.Root, u, false); err != nil {
-		return sdk.WrapError(sdk.ErrWorkflowNodeRootUpdate, "unable to update root node on workflow(%d) : %v", w.ID, err)
+		return sdk.WrapError(err, "unable to update root node on workflow(%d) : %v", w.ID, err)
 	}
 	w.RootID = w.Root.ID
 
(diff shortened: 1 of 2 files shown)


engine/api/workflow/dao.go
```diff
@@ -949,7 +949,7 @@ func Update(ctx context.Context, db gorp.SqlExecutor, store cache.Store, w *sdk.
 	w.ResetIDs()
 
 	if err := insertNode(db, store, w, w.Root, u, false); err != nil {
-		return sdk.WrapError(sdk.ErrWorkflowNodeRootUpdate, "unable to update root node on workflow(%d) : %v", w.ID, err)
+		return sdk.WrapError(err, "unable to update root node on workflow(%d) : %v", w.ID, err)
```
richardlt: I think we dont need to print (%v) the err anymore.

PR #4101, opened 2019-03-26
fix(worker): download in the current directory

### LINKED ISSUE #4095: workflow download does not download in current directory
workflow download does not download in current directory (as it says on console), but in {{.cds.workspace}}

Extract from a run of CDS with bash -ex
```
+ cd /cicd/models
+ worker download --workflow=cicd.models --number=5 '--pattern=\.tar$'
Downloading artifacts from into current directory
downloading artifact debian7.tar with tag 5-5 from workflow PCS/cicd.models on run 5...
cicd.models.5.debian9.log does not match pattern \.tar$ - skipped
cicd.models.5.log does not match pattern \.tar$ - skipped
cicd.models.5.debian7.log does not match pattern \.tar$ - skipped
+ ls -lh
total 0
+ find / -name debian7.tar
/tmp/Q0lDRA/run/debian7.tar
```

fsamin:
close #4095

### DIFF
--- a/engine/worker/cmd_download.go
+++ b/engine/worker/cmd_download.go
@@ -7,6 +7,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"os"
+	"path"
 	"regexp"
 	"strconv"
 	"sync"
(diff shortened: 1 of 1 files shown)


engine/worker/cmd_download.go
```diff
@@ -85,11 +87,13 @@ func downloadCmd(w *currentWorker) func(cmd *cobra.Command, args []string) {
 			}
 		}
 
+		wd, _ := os.Getwd()
```
bnjjj: error handling

fsamin: not useful, if the working directory is empty; it will be downloaded in the worker home directory

PR #4119, opened 2019-03-28
feat(api): add git hash short variable auto

richardlt:
1. Description
1. Related issues
1. About tests
1. Mentions

@ovh/cds


engine/hooks/poller.go
```diff
@@ -17,6 +17,7 @@ func fillPayload(pushEvent sdk.VCSPushEvent) map[string]string {
 	payload["git.author"] = pushEvent.Commit.Author.Name
 	payload["git.author.email"] = pushEvent.Commit.Author.Email
 	payload["git.branch"] = strings.TrimPrefix(strings.TrimPrefix(pushEvent.Branch.DisplayID, "refs/heads/"), "refs/tags/")
+	payload["git.hash.short"] = pushEvent.Commit.Hash[:7]
```
yesnault: length have to check to avoid a potential 'runtime error: slice bounds out of range'

PR #4139, opened 2019-04-01
fix(ui): bug actions with same name in step list, show group name in job steps

richardlt:
1. Description
1. Related issues
1. About tests
1. Mentions

@ovh/cds


ui/src/app/shared/action/action.component.ts
```diff
@@ -88,8 +83,8 @@ export class ActionComponent implements OnDestroy, OnInit {
     }
 
     ngOnInit() {
-        this.actionSub = this._actionStore.getProjectActions(this.project.key).subscribe(mapActions => {
-            this.publicActions = mapActions.valueSeq().toArray().filter((action) => action.name !== this.editableAction.name);
+        this._actionService.getAllForProject(this.project.key).subscribe(as => {
```
bnjjj: Is it relevant to put a loader here ?

richardlt: Will be nice but this list is used by a child component, so we'll have to do some changes also in children.

PR #4168, opened 2019-04-08
fix(vcs): Gitlab oauth Application callback URL

fsamin:
Warning: Users , you may need to update VCS your configuration

### DIFF
--- a/engine/vcs/types.go
+++ b/engine/vcs/types.go
@@ -86,7 +86,7 @@ var errGithubConfigurationError = fmt.Errorf("Github configuration Error")
 type GitlabServerConfiguration struct {
 	AppID            string `toml:"appId" json:"-" comment:"#######\n CDS <-> Gitlab. Documentation on https://ovh.github.io/cds/hosting/repositories-manager/gitlab/ \n#######"`
 	Secret           string `toml:"secret" json:"-"`
-	OAuthCallbackURL string `toml:"callbackUrl" json:"OAuth Application Callback URL"`
+	OAuthCallbackURL string `toml:"callbackUrl" json:"callbackUrl" default:"http://localhost:8081/repositories_manager/oauth2/callback" comment:"OAuth Application Callback URL"`
 	Status           struct {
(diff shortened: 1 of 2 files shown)


engine/vcs/vcs.go
```diff
@@ -107,7 +107,7 @@ func (s *Service) getConsumer(name string) (sdk.VCSServer, error) {
 		return gitlab.New(serverCfg.Gitlab.AppID,
 			serverCfg.Gitlab.Secret,
 			serverCfg.URL,
-			s.Cfg.API.HTTP.URL+"/repositories_manager/oauth2/callback",
```
bnjjj: Why not keep this value as default if we don't indicate `OAuthCallbackURL` ?

fsamin: Because the other value doesn't have any sens :/

engine/vcs/types.go
```diff
@@ -84,9 +84,10 @@ var errGithubConfigurationError = fmt.Errorf("Github configuration Error")
 
 // GitlabServerConfiguration represents the gitlab configuration
 type GitlabServerConfiguration struct {
-	AppID  string `toml:"appId" json:"-" comment:"#######\n CDS <-> Gitlab. Documentation on https://ovh.github.io/cds/hosting/repositories-manager/gitlab/ \n#######"`
-	Secret string `toml:"secret" json:"-"`
-	Status struct {
+	AppID            string `toml:"appId" json:"-" comment:"#######\n CDS <-> Gitlab. Documentation on https://ovh.github.io/cds/hosting/repositories-manager/gitlab/ \n#######"`
+	Secret           string `toml:"secret" json:"-"`
+	OAuthCallbackURL string `toml:"callbackUrl" json:"OAuth Application Callback URL"`
```
richardlt: Can you add a default value with http://localhost:8081/repositories_manager/oauth2/callback, to generate default config that works for dev purpose.

yesnault: json:"OAuth Application Callback URL" --> this is probably the "comment" annotation, not the json annotation :)

fsamin: fixed

PR #4187, opened 2019-04-12
chore(engine): generate cross package cover report

richardlt:
1. Description
1. Related issues
1. About tests
1. Mentions

@ovh/cds

### DIFF
--- a/engine/api/cache/redis.go
+++ b/engine/api/cache/redis.go
@@ -71,7 +71,7 @@ const retryWaitDuration = retryWait * time.Millisecond
 //Get a key from redis
 func (s *RedisStore) Get(key string, value interface{}) bool {
 	if s.Client == nil {
-		log.Error("redis> cannot get redis client")
+		log.Error("%v", sdk.WithStack(fmt.Errorf("cannot get redis client")))
 		return false
 	}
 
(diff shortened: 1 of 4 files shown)


engine/api/test/test.go
```diff
@@ -115,6 +115,7 @@ func SetupPG(t log.Logger, bootstrapFunc ...Bootstrapf) (*gorp.DbMap, cache.Stor
 		}
 	}
 
+	t.Logf("Connecting to redis store: %s", RedisHost)
```
sguiheux: remove?

engine/api/cache/redis.go
```diff
@@ -322,7 +322,7 @@ func (s *RedisStore) Subscribe(channel string) PubSub {
 // GetMessageFromSubscription from a redis PubSub
 func (s *RedisStore) GetMessageFromSubscription(c context.Context, pb PubSub) (string, error) {
 	if s.Client == nil {
-		log.Error("redis> cannot get redis client")
+		log.Error("%v", sdk.WithStack(fmt.Errorf("cannot get redis client")))
```
fsamin: i think this log appears bunch of times in test logs, won't it be to verbose with this ?

richardlt: Yes it is, used for debug purpose, removed.

PR #4188, opened 2019-04-15
fix(api): db errors

sguiheux:
1. Description
1. Related issues
1. About tests
1. Mentions

@ovh/cds


engine/api/workflow_queue.go
```diff
@@ -363,6 +363,12 @@ func (api *API) postSpawnInfosWorkflowJobHandler() service.AsynchronousHandler {
 		}
 		defer tx.Rollback()
 
+		if _, err := workflow.LoadNodeJobRun(tx, api.Cache, id); err != nil {
+			if err != sdk.ErrWorkflowNodeRunJobNotFound {
```
richardlt: Should be error is ?

PR #4189, opened 2019-04-15
fix(api): add git param with prefix for children

sguiheux:
1. Description
1. Related issues
1. About tests
1. Mentions

@ovh/cds


engine/api/workflow/process_parameters.go
```diff
@@ -177,15 +177,17 @@ func getParentParameters(w *sdk.WorkflowRun, nodeRuns []*sdk.WorkflowNodeRun) ([
 			// We inherite git variables is there is more than one repositories in the whole workflow
 			if strings.HasPrefix(param.Name, "git.") {
 				parentParams = append(parentParams, param)
+
+				// Create parent param
+				param.Name = strings.Replace(param.Name, "git.", prefix+"git.", 1)
```
richardlt: vs param.Name = prefix + param.Name ?

PR #4194, opened 2019-04-16
fix(api): import workflow with conditionnal join

sguiheux:
1. Description
1. Related issues
1. About tests
1. Mentions

@ovh/cds


engine/api/workflow_import_test.go
```diff
@@ -314,3 +713,136 @@ func Test_getWorkflowPushHandler(t *testing.T) {
 	//Check result
 	t.Logf(">>%s", rec.Body.String())
 }
+
+func Test_putWorkflowImportHandlerChocapic(t *testing.T) {
```
fsamin: what does it test ?

PR #4196, opened 2019-04-17
fix(api): fix set statement timeout option after connect

richardlt:
1. Description
1. Related issues
1. About tests
1. Mentions

@ovh/cds


ovh-cds:
CDS Report ut-engine#9255.0 ✘  
* Stage 1
  * Engine Test with DB ✘

PR #4216, opened 2019-04-24
fix(api): reset application RepositoryStrategy against ConnectionType

fsamin:
1. Description
1. Related issues
1. About tests
1. Mentions

@ovh/cds


engine/api/application/application_importer.go
```diff
@@ -18,6 +18,14 @@ func Import(db gorp.SqlExecutor, store cache.Store, proj *sdk.Project, app *sdk.
 		return sdk.WrapError(erre, "application.Import> Unable to check if application exists")
 	}
 
+	if app.RepositoryStrategy.ConnectionType == "ssh" {
```
richardlt: Can you add this code directly to the exportentities -> sdk struct convert func.

Issue #4123, opened 2019-03-29
kubernetes kubeconfig service account


BrendanBall:
To deploy to k8s the documentation says: To let CDS spawn workers on your kubernetes cluster you need to copy your kubeconfig.yaml in the current directory.
This is not a normal thing to do when using k8s. I've never had to give a deployment my local kubeconfig.yaml which contains all my clusters and their access tokens.

Should you not be creating a k8s service account with the correct permissions instead and have cds use that?

bnjjj:
I will check if it's possible. in fact I don't know if a lot of helm chart deploy a pod which also deploy himself other pods in the sames kubernetes cluster (as our hatchery do). In fact this kubeconfig.yaml is only useful to configure our kubernetes hatchery that will spawn workers (as pod) in your current kubernetes cluster.

jseguillon:
Hi. My opinion on this : yes it's a problem if you need to put admin credential in config. *obvious*

Best practice I know : 
 - set a service account with limited access
 - set a sidecar kube-proxy that will ingest service account 
 - in cds, change kube-api URLs to point on the proxy port 

Its now (more or less) usual in helm charts to add an `extraContainers` value like in this [grafana](https://github.com/helm/charts/search?utf8=%E2%9C%93&q=extraContainers) chart, to handle this case. 

Some stack discsusion here  : https://stackoverflow.com/questions/32859242/kubernetes-link-a-pod-to-kube-proxy

yesnault:
fix on master - associated to release 0.39.2

PR #4225, opened 2019-04-26
fix(hatchery): k8s hatchery uses service account instead of kubeconfig

reynn:
1. Description: Helm deploy option uses a hatchery:kubernetes which requires a kubeconfig to be present. This is generally considered bad practice from what I've seen. This will convert the k8s hatchery to using an incluster config that allows it to consume a service account token. This PR also includes updates to helm charts to configure this. I've also attached a dockerfile for dev purposes that builds just the minimum required for faster local dev.
1. Related issues: #4123
1. About tests
1. Mentions

@ovh/cds


yesnault:
Thank you a lot! 
There is probably missing some documentation update on https://github.com/ovh/cds/blob/master/contrib/helm/cds/README.md, we'll have a look at this next week

yesnault:
In fact, there is two things:
 - Hatchery k8s can be used by a CDS API outside k8s, so, without using helm
 - Hatchery k8s can be used by a CDS API inside k8s, with helm deploy for example

reynn:
I reverted some of the code so that you can use a kubeconfig or service account, also updated the helm readme to remove references to copying the kubeconfig into the helm directory.

richardlt:
Why was the ui/package.json updated in this PR ?

PR #4303, opened 2019-05-17
fix(hatchery/swarm): improve some logs

yesnault:
hatchery swarm: avoid error log if a container is already under deletion


engine/api/worker.go
```diff
@@ -90,7 +90,7 @@ func (api *API) disableWorkerHandler() service.Handler {
 			return sdk.WrapError(sdk.ErrNotFound, "disabledWorkerHandler> Cannot load worker %s", id)
 		}
 
-		if wor.Status == sdk.StatusBuilding {
+		if wor.Status == sdk.StatusBuilding && getHatchery(ctx) == nil {
```
yesnault:
```suggestion
		if wor.Status == sdk.StatusBuilding {
```

PR #4321, opened 2019-05-22
feat(api, cdsctl): cdsctl project / workflow favorite list

yesnault:
two new commands to allow the vscode plugin list favorites projects and workflows


yesnault:
@fsamin yes, fixed

yesnault:
I prefer queryParams instead of using getNavbarHandler. getNavbarHandler is really linked to the webUI with different content for an admin user containing all data vs standard user. getNavbarHandler for an admin user seems really to big.

PR #4331, opened 2019-05-24
fix(ui): missing shared service import generates js error

richardlt:
1. Description
1. Related issues
1. About tests
1. Mentions

@ovh/cds


ui/src/app/shared/parameter/value/parameter.value.component.ts
```diff
@@ -76,7 +77,8 @@ export class ParameterValueComponent implements OnInit, AfterViewChecked {
 
     constructor(
         private _repoManagerService: RepoManagerService,
-        private _theme: ThemeStore
+        private _theme: ThemeStore,
+        public _sharedService: SharedService // use in html
```
bnjjj:
```suggestion
        public _sharedService: SharedService // used in html
```

PR #4334, opened 2019-05-25
fix(vcs/gitlab): wrong callbackurl variable in config.toml

gregoiremahe:
1. Description
Default value when you generate CDS configuration with cds-engine config new is wrong for the callback URL
CDS + GitLab link doesn't work with : 
 
`[vcs.servers.Gitlab.gitlab]
        appId = "xxxx"
        callbackUrl = ""
        disablePolling = false
        disableWebHooks = false
        # proxyWebhook = ""
        secret = "xxxx"`

but works with
`[vcs.servers.Gitlab.gitlab]
        appId = "xxxx"
        OAuthCallbackURL = ""
        disablePolling = false
        disableWebHooks = false
        # proxyWebhook = ""
        secret = "xxxx"`
 
2. Related issues
Not created, but related to https://github.com/ovh/cds/pull/4168

3. About tests
It's ok on our side, but you can still test with the container located at docker.io/asforge/cds-engine:0.39.3-60

4. Mentions
@ovh/cds

### DIFF
--- a/docs/content/docs/integrations/gitlab.md
+++ b/docs/content/docs/integrations/gitlab.md
@@ -63,6 +63,9 @@ Set value to `appId` and `secret`
         # proxyWebhook = ""
         secret = "xxxx"
 
+        # OAuth Application Callback URL
+        OAuthCallbackURL = "http://your-cds-api:8081/repositories_manager/oauth2/callback"
+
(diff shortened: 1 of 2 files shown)


richardlt:
Thanks, I already fix this in #4271 ;) . This will be in the next release and the key will still be "callbackUrl".

PR #4340, opened 2019-05-27
fix(contrib/helm): replace --decode by -d to be compatible with more distribs / os

gregoiremahe:
replace --decode by -d to be compatible with more distribs / os

1. Description
base64 --decode is not working on alpine, whereas base64 -d is working on alpine, debian, darwin ... 

2. Related issues
None

3. About tests
tested on debian, macos, alpine

4. Mentions
@bnjjj 
@ovh/cds


gregoiremahe:
> can you rebase this PR with master?

done

PR #4345, opened 2019-06-02
WIP: feat(metrics): refactor the ElasticSearch service to be generic for further options

reynn:
If this is acceptable I intend to add support for additional metric providers starting with DataDog

1. Description: The goal of this PR is to start moving the `elasticsearch` service to be more generic. This will become a `metrics` service that works similar to the `vcs` service and will allow admins to use one service that can push to several metric providers. Old version of the ElasticSearch service will not work on ElasticSearch 6 and above due to mapping types outlined here: https://www.elastic.co/guide/en/elasticsearch/reference/6.0/breaking-changes-6.0.html.
1. About tests: I've tried this on a fresh install of ES 6.8.0 with success, I have not been able to test these changes with an old index from ES 5.x.
1. Mentions

Tasks:

- [x] Update Config struct in `engine/api/api.go` and `engine/config.go`
- [x] Rename package `engine/elasticsearch` to `engine/metricsservice`
- [x] Update function/method names that explicitly used `elasticsearch`
- [x] Change `services.TypeElasticsearch`

- [ ] Bug: Timeline and Event Heatmap on home page not working properly
- [ ] Determine desire for backwards compatibility or come up with a migration strategy

@ovh/cds


yesnault:
Hello @reynn 

Thank you for this pull request. After a debrief on this service:
 - we will create two services: one for event, another for metrics (release 0.42.0)
 - we will delete the existing service elasticsearch (release 0.43.0) - no backward compatibility planned on CDS Conf with the current elasticsearch service, but users will be able to keep the existing backend.

The event service will manage "event data" with the actual GET,POST on /events
The metrics service will manage the actual GET/POST on /metrics.

The backend elasticsearch will be implemented for both as today. This will let the choice to implement datadog, ovh metrics, other backend on the adequate service. 

After this new service, the timeline will get document from event service, a new index is planned to store only one document for each workflow run, and not many as today.

Thank you again for this PR, I'll close it, the work is in progress about the split of this service.

yesnault:
The "split" will be develop after https://github.com/ovh/cds/pull/4290 to avoid too many merge conflicts.

PR #4363, opened 2019-06-07
feat(api,ui): add support of badge generation

### LINKED ISSUE #575: feat (contrib): cds badges

bnjjj:

1. Description
1. Related issues
close #575
1. About tests
1. Mentions

@ovh/cds

### DIFF
--- a/contrib/uservices/badge/src/kafka.rs
+++ b/contrib/uservices/badge/src/kafka.rs
@@ -78,8 +78,7 @@ impl Actor for KafkaConsumerActor {
                         continue;
                     }
                     match serde_json::from_slice::<Event>(payload.unwrap()) {
-                        Ok(ref event) if event.type_event == "sdk.EventRunWorkflow" => {
-                            // TODO: add a boolean to check if we want to generate badge
+                        Ok(ref event) if event.type_event == "sdk.EventRunWorkflow" && event.generate_badge => {
                             let mut branch = None;
                             if let Some(ref tags) = &event.tag {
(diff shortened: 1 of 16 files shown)


fsamin:
Here is a suggestion, assuming :
* that badge management is only the business of the badge service
* users should be able to create badge to put them on some html pages and readme
* only internal users can create badges
* external users should not be able to predict badges URL to avoid leaking data

The badge uservices is fed by the kafka topic and can generate badge on all workflows.
On badge creation, the badge uservices doesn't store anything, it only give the user an URL made of a JWS (https://badge.cds.ovh.net/<JWS>).

The JWS is signed by the uservices and contains: project and workflow name.

When the uservice is called for a badge: it checks the signature of the JWS, if it ok: it renders the badge for the given project and name.

To create the badge we deploy a single internal web page. Which gives the users ability to create badge.
The web page can be a "CDS provider" to be able list all projects and workflows from a users point of views. To retrieve the user: we can get it from headers because the web page would be deployed on static gateway.
Then the web page calls an API to sign a piece of data with a private key.

The badge uservice checks the signature against the API public key.

richardlt:
I agree with @fsamin suggestion, URL should not be predictable.
However the first idea from this PR was to store an information to enable/disable the badge and this is also something useful if we want to be sure that the status or logs (in the future) are no longer visible.
Also because we need to check user permissions to allow url generation, this should relies on the API and the user session. I'm not sure that the badge uservice should handle that part.

PR #4400, opened 2019-06-26
feat(api): manually delete 1 build from history #3782

### LINKED ISSUE #3782: feat: manually delete 1 build from history
Hi,

Is there any way to manually remove 1 build from history without removing all the bulid before ?
The idea is to delete build log/artifacts/everything from history if there is a leak on it (clear password, key, etc…)
Thanks

bnjjj:

1. Description
1. Related issues
close #3782 
1. About tests
1. Mentions

@ovh/cds


ovh-cds:
CDS Report ut-engine#10029.0 ✘  
* Stage 1
  * Engine Test with DB ✘

engine/api/workflow/dao_run.go
```diff
@@ -41,6 +41,7 @@ type LoadRunOptions struct {
 	WithTests               bool
 	WithLightTests          bool
 	WithVulnerabilities     bool
+	WithoutDeleted          bool
```
richardlt: To prevent returns deleted runs, maybe we should have a WithDeleted flag instead of Without. So the default case will be to return only not deleted runs. No ?

engine/api/workflow/dao_run.go
```diff
@@ -469,6 +473,9 @@ func loadRun(db gorp.SqlExecutor, loadOpts LoadRunOptions, query string, args ..
 		return nil, sdk.WrapError(err, "Unable to load workflow run. query:%s args:%v", query, args)
 	}
 	wr := sdk.WorkflowRun(*runDB)
+	if loadOpts.WithoutDeleted && wr.ToDelete {
+		return nil, sdk.ErrWorkflowNotFound
```
richardlt:
```suggestion
		return nil, sdk.WithStack(sdk.ErrWorkflowNotFound)
```

PR #4473, opened 2019-07-19
fix(api,repositories): use tag with as code workflow

sguiheux:
1. Description
1. Related issues
1. About tests
1. Mentions

@ovh/cds

### DIFF
--- a/engine/api/workflow/process.go
+++ b/engine/api/workflow/process.go
@@ -12,14 +12,21 @@ import (
 func setValuesGitInBuildParameters(run *sdk.WorkflowNodeRun, vcsInfos vcsInfos) {
 	if run.ApplicationID != 0 {
 		run.VCSRepository = vcsInfos.Repository
-		run.VCSBranch = vcsInfos.Branch
+		if vcsInfos.Tag == "" {
+			run.VCSBranch = vcsInfos.Branch
+		}
+
 		run.VCSTag = vcsInfos.Tag
 		run.VCSHash = vcsInfos.Hash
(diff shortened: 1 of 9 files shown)


engine/api/workflow/repository.go
```diff
@@ -98,7 +98,7 @@ func extractWorkflow(ctx context.Context, db *gorp.DbMap, store cache.Store, p *
 		RepositoryStrategy: ope.RepositoryStrategy,
 		Branch:             ope.Setup.Checkout.Branch,
 		FromRepository:     ope.RepositoryInfo.FetchURL,
-		IsDefaultBranch:    ope.Setup.Checkout.Branch == ope.RepositoryInfo.DefaultBranch,
+		IsDefaultBranch:    ope.Setup.Checkout.Tag == "" && ope.Setup.Checkout.Branch == ope.RepositoryInfo.DefaultBranch,
```
richardlt: Branch is not empty when it's a tag ?

engine/api/workflow/process.go
```diff
@@ -12,14 +12,21 @@ import (
 func setValuesGitInBuildParameters(run *sdk.WorkflowNodeRun, vcsInfos vcsInfos) {
 	if run.ApplicationID != 0 {
 		run.VCSRepository = vcsInfos.Repository
-		run.VCSBranch = vcsInfos.Branch
+		if vcsInfos.Tag == "" {
```
richardlt: Should we set run.VCSTag Only when vcsInfos.Tag != "" ?

PR #4481, opened 2019-07-23
Update ldapclient.go

### LINKED ISSUE #4479: Ldap authentication
I couldn't authenticate with LDAP , I had this Error `[WARN] LDAP> Bind error elabouti LDAP Result Code 32 "No Such Object" `  , because while authenticating with LDAP server ,CDS sent the LdapBind  request with just the username , but it should send firstly the Ldapsearch request  with the username to get the DN then send the LdapBind with this DN to authenticate

ELABOUTISamiha:
fixes #4479 Ldap auth

Before this modifications I couldn't authenticate with lLDAP , because CDS sent wrong requests , so it couldn't find users in my LDAP server  

@ovh/cds

### DIFF
--- a/engine/api/auth/ldapclient.go
+++ b/engine/api/auth/ldapclient.go
@@ -231,12 +231,10 @@ func (c *LDAPClient) BindDN(dn, password string) error {
 
   if err := c.conn.Bind(dn, password); err != nil {
     if shoudRetry(err) {
-      err = c.openLDAP(c.conf)
-      if err != nil {
+      if err = c.openLDAP(c.conf); err != nil {
         return err
       }
-      err = c.conn.Bind(dn, password)
-      if err != nil {
+      if err = c.conn.Bind(dn, password); err !=nil {
         return err
       }
     } else {
@@ -399,15 +397,17 @@ func (c *LDAPClient) Authentify(username, password string) (bool, error) {
   }
   
 	//Bind user
-	if err := c.BindDN(r[0].DN, password); err != nil {
-		log.Warning("LDAP> Bind error %s %s", username, err)
+  if r != nil {
+	  if err := c.BindDN(r[0].DN, password); err != nil {
+		  log.Warning("LDAP> Bind error %s %s", username, err)
 
-		if !isCredentialError(err) {
-			return false, err
-		}
-		//Try local auth
-		return c.local.Authentify(username, password)
-	}
+		  if !isCredentialError(err) {
+			  return false, err
+		  }
+		  //Try local auth
+		  return c.local.Authentify(username, password)
+	  }
+  }
 
 	log.Debug("LDAP> Bind successful %s", username)
 
--- a/engine/api/auth/ldapclient.go
+++ b/engine/api/auth/ldapclient.go
@@ -225,23 +225,6 @@ func (c *LDAPClient) Bind(username, password string) error {
 	return nil
 }
 
-// BindDN
-func (c *LDAPClient) BindDN(dn, password string) error {
-  log.Debug("LDAP> Bind DN %s", dn)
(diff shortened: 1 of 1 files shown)


engine/api/auth/ldapclient.go
```diff
@@ -370,8 +392,14 @@ func (c *LDAPClient) searchAndInsertOrUpdateUser(db gorp.SqlExecutor, username s
...
 func (c *LDAPClient) Authentify(username, password string) (bool, error) {
+  // Search user
+	r, err := c.Search("(&(uid="+username+"))")
+	if err != nil {
+	  return false, nil
+  }
+  
 	//Bind user
-	if err := c.Bind(username, password); err != nil {
+	if err := c.BindDN(r[0].DN, password); err != nil {
```
richardlt: Should we check that r is not empty ?

engine/api/auth/ldapclient.go
```diff
@@ -370,8 +392,14 @@ func (c *LDAPClient) searchAndInsertOrUpdateUser(db gorp.SqlExecutor, username s
 
 //Authentify check username and password
 func (c *LDAPClient) Authentify(username, password string) (bool, error) {
+  // Search user
+	r, err := c.Search("(&(uid="+username+"))")
```
richardlt: Need gofmt

engine/api/auth/ldapclient.go
```diff
@@ -225,6 +225,28 @@ func (c *LDAPClient) Bind(username, password string) error {
 	return nil
 }
 
+// BindDN
+func (c *LDAPClient) BindDN(dn, password string) error {
+  log.Debug("LDAP> Bind DN %s", dn)
+
+  if err := c.conn.Bind(dn, password); err != nil {
+    if shoudRetry(err) {
+      err = c.openLDAP(c.conf)
```
richardlt: One line err check: if err := c.openldap....

engine/api/auth/ldapclient.go
```diff
@@ -225,6 +225,28 @@ func (c *LDAPClient) Bind(username, password string) error {
...
+      if err != nil {
+        return err
+      }
+      err = c.conn.Bind(dn, password)
```
richardlt: Same one line err

engine/api/auth/ldapclient.go
```diff
@@ -225,6 +225,28 @@ func (c *LDAPClient) Bind(username, password string) error {
...
+      if err != nil {
+        return err
+      }
+    } else {
```
richardlt: We can do early return: if !shouldretry

engine/api/auth/ldapclient.go
ELABOUTISamiha: I think it isn't that necessary to check the same error openldap in the ldapbind because we check it in the ldapsearch

engine/api/auth/ldapclient.go
ELABOUTISamiha: Yeah, I just tried to keep the same structure of the other function  func (c *LDAPClient) Bind(username, password..

engine/api/auth/ldapclient.go
```diff
@@ -370,15 +387,23 @@ func (c *LDAPClient) searchAndInsertOrUpdateUser(db gorp.SqlExecutor, username s
...
+	// Search user
+	r, err := c.Search("(&(uid=" + username + "))")
+	if err != nil {
+		return false, nil
+	}
+
 	//Bind user
-	if err := c.Bind(username, password); err != nil {
-		log.Warning("LDAP> Bind error %s %s", username, err)
+	if r != nil && len(r) == 1 {
```
golangcibot: S1009: should omit nil check; len() for nil slices is defined as zero (from `gosimple`)

PR #4555, opened 2019-08-22
fix(hooks): init kafka / rabbitmq task

yesnault:
this will avoid to delete unprocessed tasks on func deleteTaskExecutionsRoutine


yesnault:
dequeueTaskExecutions must be set on func `dequeueTaskExecutions`

PR #4584, opened 2019-09-04
fix(ui): select in node condition display uknown variable #4569

### LINKED ISSUE #4569: fix(ui): select in node condition should display unknown keys

bnjjj:

1. Description
1. Related issues
close #4569 
1. About tests
1. Mentions

@ovh/cds


ui/src/app/shared/conditions/conditions.html
```diff
@@ -35,10 +35,11 @@
                     <tr *ngFor="let c of conditions?.plain; let i = index">
                         <td class="middle-aligned">
                             <div class="field">
-                                <ng-container *ngIf="conditionNames">
+                                <ng-container *ngIf="conditionNames && !c.unknownVariable">
```
richardlt: Is it possible to set a value that is not in the list ? when never run

bnjjj: old and deprecated field. I deleted

PR #4586, opened 2019-09-05
fix(hooks): delete the branch deletion when this branch is pushed again #4553

### LINKED ISSUE #4553: fix(hooks): when a branch is deleted and pushed again it doesn't delete the delete run task

bnjjj:

1. Description
1. Related issues
close #4553 
1. About tests
1. Mentions

@ovh/cds

### DIFF
--- a/engine/hooks/branch_deletion.go
+++ b/engine/hooks/branch_deletion.go
@@ -15,3 +15,21 @@ func (s *Service) doBranchDeletionTaskExecution(t *sdk.TaskExecution) (*sdk.Work
 
 	return nil, sdk.WrapError(err, "cannot mark to delete workflow runs")
 }
+
+func (s *Service) stopBranchDeletionTask(branch string) error {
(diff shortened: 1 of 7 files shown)


engine/hooks/branch_deletion.go
```diff
@@ -15,3 +15,21 @@ func (s *Service) doBranchDeletionTaskExecution(t *sdk.TaskExecution) (*sdk.Work
 
 	return nil, sdk.WrapError(err, "cannot mark to delete workflow runs")
 }
+
+func (s *Service) stopBranchDeletionTask(branch string) error {
+	keys, err := s.Dao.FindAllKeysMatchingPattern(branch + "*")
```
yesnault: could you add a check to check if branch is not empty (to avoid drop everything)

bnjjj: Good idea ! Haha

PR #4605, opened 2019-09-13
Fix #4604

### LINKED ISSUE #4604: fix(worker) worker tmpl use absolute path instead of relative
The `worker tmpl` command uses an absolute path resolution from the `{{.cds.workspace}}`.

So if we are in a child directory, we need to redefine all the path

XciD:
Add current directory to worker tmpl command

@ovh/cds

### DIFF
--- a/engine/worker/cmd_tmpl.go
+++ b/engine/worker/cmd_tmpl.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"path/filepath"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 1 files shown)


XciD:
Updated

engine/worker/cmd_tmpl.go
```diff
@@ -98,6 +108,14 @@ func tmplCmd(w *currentWorker) func(cmd *cobra.Command, args []string) {
 	}
 }
 
+func getAbsoluteDir(arg string, currentDir string) string {
```
yesnault: With early return:
```go
func getAbsoluteDir(arg string, currentDir string) string {
	if strings.HasSuffix(arg, string(filepath.Separator)) {
		return arg
	}
	return filepath.Join(currentDir, arg)
}
```

--HasPrefix --> it's HasPrefix, no? :)

XciD: You are right ;)

PR #4718, opened 2019-10-26
fix(api): run wf with mutex, from a hook

### LINKED ISSUE #4715: fix(api): child workflow with first pipeline with mutex
[2019-10-25T14:18:32] Workflow upload-to#147.0 has been started
[2019-10-25T14:18:32] The pipeline upload-to is waiting while it's running on another run
[2019-10-25T14:18:32] Run conditions aren't ok.

--> the parent wf is still in building
--> the child is stopped (not launched), it should be waiting, then building

yesnault:
and various fixes on some log

close #4715


yesnault:
this PR avoid to have two node run with a mutex running in the same time too

Issue #4775, opened 2019-11-22
chore: package plugin-kubernetes with release


yesnault:
source: https://gitter.im/ovh-cds/Lobby?at=5dd6bc095b3db9548ac7ae9f

yesnault:
fixed, feel free to reopen if needed

PR #4777, opened 2019-11-22
#4775 Fix build issue in makefile for kubernetes deployment plugin

louis-gounot:
#4775 Fix build issue in makefile for kubernetes deployment plugin

@ovh/cds


louis-gounot:
Issue I tried to fix is #4778 instead.
See #4779 for PR.

PR #4785, opened 2019-11-27
fix: do not allow additionalProperties

sguiheux:
1. Description
1. Related issues
1. About tests
1. Mentions

@ovh/cds

### DIFF
--- a/cli/cdsctl/tools.go
+++ b/cli/cdsctl/tools.go
@@ -107,7 +107,7 @@ func toolsYamlSchemaRun(v cli.Values) error {
 	results := make([]string, len(types))
 	for i := range types {
 		r := jsonschema.Reflector{
-			AllowAdditionalProperties:  true,
+			AllowAdditionalProperties:  false,
 			RequiredFromJSONSchemaTags: true,
 		}
 		sch := r.ReflectFromType(types[i])


ovh-cds:
CDS Report cds-package#11487.0 ✘  
* build-package
  * Debian Engine Package ✘  
  * Docker Engine Package ✔ 
  * Package UI ✔

cli/cdsctl/tools.go
```diff
@@ -107,7 +107,7 @@ func toolsYamlSchemaRun(v cli.Values) error {
 	results := make([]string, len(types))
 	for i := range types {
 		r := jsonschema.Reflector{
-			AllowAdditionalProperties:  true,
+			AllowAdditionalProperties:  false,
```
richardlt: This was useful to ignore error in yaml when a unknown entry is set. This was set to true to ignore error for custom action where custom names are unknown keys in yaml.

richardlt: Why do we need to change it ?

sguiheux: to be able to validate a yaml. To avoid that people enter a key that doesn ’t match the schema.
Today I can type foo: bar in a workflow yaml file, and it’s ok

richardlt:
Closed because of #4872

Issue #4800, opened 2019-12-09
Error JS when removing a pipeline from a workflow


sguiheux:
```
console.js.pre-build-optimizer.js:35 ERROR TypeError: Cannot read property 'name' of undefined
    at Object.updateDirectives (menu.edit.node.html:46)
```

Issue #4762, opened 2019-11-13
{{.url}} in payload is taken in GitClone action


blaryjp:
Hello,

If you put in your payload "{ "url": "toto.com" }" , this variabla will be taken in GitClone as the Git URL, even if you have hard coded in URL.

Thx

bnjjj:
fixed

PR #4823, opened 2019-12-13
fix(worker,ui): error JS #4800 and build params and action params #4762

bnjjj:

1. Description
1. Related issues
1. About tests
1. Mentions

@ovh/cds


bnjjj:
tested on different workflow with integrations tests

bnjjj:
fixed in other PR

PR #4855, opened 2020-01-02
fix(ui,api): do not load condition names when it's not needed #4846

### LINKED ISSUE #4846: fix(ui): cannot get condition on workflow run node
Nil pointer when getting condition on a node.
Occurs because the handler use the workflow data of the last run. 
If previous run uses a workflow version that is not the one used by the last run the node id cannot be found.

https://github.com/ovh/cds/blob/69a28f0814357b204ae0024d61cc8fb47b217459/engine/api/workflow_trigger.go#L16

bnjjj:

1. Description
1. Related issues
close #4846 
1. About tests
1. Mentions

@ovh/cds

### DIFF
--- a/ui/src/app/shared/workflow/wizard/conditions/wizard.conditions.component.ts
+++ b/ui/src/app/shared/workflow/wizard/conditions/wizard.conditions.component.ts
@@ -118,26 +118,28 @@ export class WorkflowWizardNodeConditionComponent extends Table<WorkflowNodeCond
             }
         });
 
-        if (this.node) {
-            this._workflowService.getTriggerCondition(this.project.key, this.workflow.name, this.node.id)
-                .pipe(
-                    first(),
-                    finalize(() => {
-                        this.loadingConditions = false;
-                        this._cd.markForCheck();
-                    })
-                )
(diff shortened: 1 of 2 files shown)


ui/src/app/shared/workflow/wizard/conditions/wizard.conditions.component.ts
```diff
@@ -118,26 +118,28 @@ export class WorkflowWizardNodeConditionComponent extends Table<WorkflowNodeCond
...
-            this._workflowService.getTriggerHookCondition(this.project.key, this.workflow.name)
-                .pipe(
-                    first(),
-                    finalize(() => {
-                        this.loadingConditions = false;
-                        this._cd.markForCheck();
-                    })
-                )
-                .subscribe(wtc => this.triggerConditions = wtc);
+        if (!this.readonly) {
```
sguiheux: No side effect on a workflow run, when we want to see run condition ?

bnjjj: no, only used to fill form

PR #4878, opened 2020-01-11
fix(cdsctl): cdsctl version

yesnault:
display
{"version":"snapshot","architecture":"amd64","os":"darwin","git_hash":"","build_time":"","db_migrate":"","keychain":true}

instead of:
{"keychain":true,"version":"CDS  version:snapshot os:darwin architecture:amd64 git.hash: build.time: db.migrate:"}

 - do not display version of api in cdsctl version
 - fix Makefile docker-compose, to let user specifiy the docker image

### DIFF
--- a/cli/cdsctl/version.go
+++ b/cli/cdsctl/version.go
@@ -4,7 +4,6 @@ import (
 	"encoding/json"
 	"fmt"
 
-	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	yaml "gopkg.in/yaml.v2"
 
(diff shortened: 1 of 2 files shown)


cli/cdsctl/version.go
```diff
@@ -30,43 +29,30 @@ func version() *cobra.Command {
...
 func versionRun(v cli.Values) error {
-	var apiVersion *sdk.Version
-	var err error
-	if client != nil {
-		apiVersion, err = client.Version()
-	} else {
-		err = errors.New("no configuration file found")
+	format := v.GetString("format")
+	if format == "" {
+		fmt.Println(sdk.VersionString())
```
fsamin: with the default "" format the string: `CDS %s version:%s os:%s architecture:%s git.hash:%s build.time:%s db.migrate:%s` is not very clear for the user.

i would prefer to display:
* version of the CLI (with architecture, keychain enable, etc...)
* URL version of the connected API (if available)

something like `docker version`

yesnault: done
```bash
$ cdsctl version --format json
{"cdsctl":{"version":"snapshot","architecture":"amd64","os":"darwin","git_hash":"","build_time":"","keychain":true},"api":{"version":"snapshot","architecture":"amd64","os":"darwin","git_hash":"","build_time":""}}
 
$ cdsctl version
cdsctl:
  version: snapshot
  architecture: amd64
  os: darwin
  git_hash: ""
  build_time: ""
  keychain: true
api:
  version: snapshot
  architecture: amd64
  os: darwin
  git_hash: ""
  build_time: ""

PR #4916, opened 2020-01-23
fix(api): get projects from provider with username

fsamin:
1. Description
1. Related issues
1. About tests
1. Mentions

@ovh/cds

### DIFF
--- a/engine/api/project.go
+++ b/engine/api/project.go
@@ -88,7 +88,14 @@ func (api *API) getProjectsHandler_FilterByRepo(ctx context.Context, w http.Resp
 		return err
 	}
 	for i := range projects {
+		if isAdmin(ctx) {
+			projects[i].Permissions = sdk.Permissions{Readable: true, Writable: true, Executable: true}
+			continue
+		}
 		projects[i].Permissions = perms[projects[i].Key]
+		if isMaintainer(ctx) {
+			projects[i].Permissions.Readable = true
+		}
 	}
 
 	if strings.ToUpper(withPermissions) == "W" {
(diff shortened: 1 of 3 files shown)


engine/api/project.go
```diff
@@ -83,74 +163,22 @@ func (api *API) getProjectsHandler() service.Handler {
...
+		var groupIDs []int64
+		if requestedUser == nil {
+			groupIDs = getAPIConsumer(ctx).GetGroupIDs()
+		} else {
+			groupIDs = requestedUser.GetGroupIDs()
+		}
+
 		pKeys := projects.Keys()
-		perms, err := permission.LoadProjectMaxLevelPermission(ctx, api.mustDB(), pKeys, getAPIConsumer(ctx).GetGroupIDs())
+		perms, err := permission.LoadProjectMaxLevelPermission(ctx, api.mustDB(), pKeys, groupIDs)
```
richardlt: Looks like this code is useful only if withPermission, also maintainer an admin can have no permission on project based on groups but we should add read permission for maintainer and all permissions for admin.

engine/api/project.go
```diff
@@ -27,12 +27,93 @@ import (
...
+		}
+	} else {
+		projects, err = project.LoadAllByRepoAndGroupIDs(ctx, api.mustDB(), api.Cache, getAPIConsumer(ctx).GetGroupIDs(), filterByRepo, opts...)
+		if err != nil {
+			return err
+		}
+	}
+
+	pKeys := projects.Keys()
+	perms, err := permission.LoadProjectMaxLevelPermission(ctx, api.mustDB(), pKeys, getAPIConsumer(ctx).GetGroupIDs())
```
richardlt: Same comments

engine/api/project.go
fsamin: We always need permission. BWT i fixed it for admin, and maintainer permissions

engine/api/project.go
fsamin: fixed

PR #4933, opened 2020-01-30
fix(ui): cannot add a worker model without required info #4816

### LINKED ISSUE #4816: fix(ui): add worker model with a wrong value
to avoid a 400 from API, add a check on ui only to show the missing value on ui, as the `Shell command`

bnjjj:

1. Description
1. Related issues
close #4816 
1. About tests
1. Mentions

@ovh/cds

### DIFF
--- a/ui/src/app/views/settings/worker-model/form/worker-model.form.component.ts
+++ b/ui/src/app/views/settings/worker-model/form/worker-model.form.component.ts
@@ -1,12 +1,4 @@
-import {
-    ChangeDetectionStrategy, ChangeDetectorRef,
-    Component,
-    EventEmitter,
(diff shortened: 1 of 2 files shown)


ui/src/app/views/settings/worker-model/form/worker-model.form.component.ts
```diff
@@ -109,6 +101,37 @@ pattern_name: basic_unix`;
             .subscribe((wmStr) => this.workerModelAsCode = wmStr);
     }
 
+    canSave(): boolean {
```
sguiheux: Does we need something to indicate mandatory field or something for users?

PR #4964, opened 2020-02-07
fix(api): delete corrupted sessions

fsamin:


engine/api/authentication/session.go
```diff
@@ -74,6 +74,7 @@ func SessionCleaner(ctx context.Context, dbFunc func() *gorp.DbMap) {
 	log.Info(ctx, "Initializing session cleaner...")
 	db := dbFunc()
 	tick := time.NewTicker(10 * time.Second)
+	tickCorruped := time.NewTicker(12 * time.Hour)
```
sguiheux: add defer or // TODELEtE

fsamin: done

PR #4979, opened 2020-02-14
fix(api): ut for usernames selected to send notifications

yesnault:

### DIFF
--- a/engine/api/notification/permission_test.go
+++ b/engine/api/notification/permission_test.go
@@ -1,0 +1,63 @@
+package notification
+
+import (
+	"context"
+	"testing"
+
+	"github.com/alecthomas/assert"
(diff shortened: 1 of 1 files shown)


engine/api/notification/permission_test.go
```diff
@@ -0,0 +1,63 @@
...
+	}), "unable to insert user in group")
+
+	assert.NoError(t, project.Update(db, cache, proj))
+
+	group.DefaultGroup = g1
+
+	userList, err := projectPermissionUserIDs(context.Background(), db, cache, proj.ID, sdk.PermissionRead)
+	assert.NoError(t, err)
+	assert.NotEmpty(t, userList)
+	assert.Equal(t, 1, len(userList))
```
fsamin: can you check this is the right userID  ?

Issue #4946, opened 2020-02-03
fix(scheduler): scheduler executed even if not present in as-code file


blaryjp:
Hello,

I 've schedulers which are executed, even if it's no more in as-code file (was added and then deleted).

See "run-testcafe" internaly ;)

Thx!

PR #4992, opened 2020-02-19
fix(sdk): dont try to run workflow from hook if the hook doesn't exist

fsamin:


fsamin:
closes #4946

PR #5076, opened 2020-03-24
fix(api): interpolate notif event

yesnault:


engine/api/notification/user.go
```diff
@@ -29,6 +29,7 @@ func GetUserWorkflowEvents(ctx context.Context, db gorp.SqlExecutor, store cache
 
 	//Compute notification
 	params := map[string]string{}
+	log.Info(ctx, "############## nr.BuildParameters: %+v", nr.BuildParameters)
```
richardlt: Debug ?

yesnault: yep

PR #5109, opened 2020-04-07
fix: start workflow with a fork

sguiheux:
1. Description
1. Related issues
1. About tests
1. Mentions

@ovh/cds


engine/api/workflow/dao.go
```diff
@@ -1183,6 +1183,9 @@ func checkHooks(db gorp.SqlExecutor, w *sdk.Workflow, n *sdk.Node) error {
 			w.HookModels[hm.ID] = *hm
 			h.HookModelID = hm.ID
 		}
+		if h.HookModelName == sdk.RepositoryWebHookModelName && (n.Context == nil || n.Context.ApplicationID == 0) {
```
fsamin, line 1186: there is a risk that we will be unable to load some existing workflows, we have to make a migration is any

PR #5131, opened 2020-04-15
feat(hooks): add logs and balance in/out in status

fsamin:

### DIFF
--- a/engine/hooks/scheduler.go
+++ b/engine/hooks/scheduler.go
@@ -17,28 +17,28 @@ func (s *Service) runScheduler(c context.Context) error {
 
 	go func() {
 		if err := s.dequeueTaskExecutions(ctx); err != nil {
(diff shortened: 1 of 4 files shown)


engine/hooks/scheduler.go
```diff
@@ -336,7 +340,10 @@ func (s *Service) dequeueTaskExecutions(ctx context.Context) error {
 
 		//Start (or restart) the task
 		if restartTask {
-			_, _ = s.startTask(ctx, task)
+			_, err := s.startTask(ctx, task)
+			if err != nil {
+				log.Error(ctx, "dequeueTaskExecutions> unable to restart the task %+v after execution: %v", err)
```
golangcibot, line 345: printf: Error format %v reads arg #2, but call has 1 arg (from `govet`)

PR #5134, opened 2020-04-17
feat(api): lock update ascode with template, workflow pull with template

richardlt:
1. Description
Improve errors returned by repositories, save only user part in operation.
Lock ascode update for a workflow generated from a template.
Fix template apply form in UI with key param type.
Workflow pull now take care of workflow generated from a template and return only template data.

1. Related issues
1. About tests
1. Mentions

@ovh/cds


ui/src/app/service/authentication/logout.interceptor.ts
```diff
@@ -33,13 +33,19 @@ export class LogoutInterceptor implements HttpInterceptor {
...
                         // error formatted from CDS API
-                        if (e.error && e.error.message) {
-                            this._toast.error(e.statusText, e.error.message);
-                        } else {
-                            try {
-                                this._toast.error(e.statusText, JSON.parse(e.message));
-                            } catch (e) {
+                        if (e.error) {
+                            if (e.error.message) {
+                                this._toast.error(e.statusText, e.error.message);
```
fsamin, line 40: this could be useful to display the requestID

PR #5182, opened 2020-05-14
fix(api): compute usage of worker model

yesnault:


yesnault:
cf. 5183

PR #5274, opened 2020-06-23
feat(venom plugin): add stop-on-failure parameter to venom plugin

storagebits:
1. Description
stop-on-failure venom option was not available in venom plugin parameters

1. Related issues
None

1. About tests
Have tested parameter behavior is OK on my CDS instance

1. Mentions

@ovh/cds
Signed-off-by: Francois Lesage francois.lesage@gmail.com

### DIFF
--- a/contrib/grpcplugins/action/plugin-venom/main.go
+++ b/contrib/grpcplugins/action/plugin-venom/main.go
@@ -76,6 +76,15 @@ func (actPlugin *venomActionPlugin) Run(ctx context.Context, q *actionplugin.Act
 		}, nil
 	}
 
+	stopOnFailure := false
+	if stopOnFailureStr != "" {
+		var errb error
+		stopOnFailure, errb = strconv.ParseBool(stopOnFailureStr)
+		if err != nil {
+			return actionplugin.Fail("Error parsing stopOnFailure value : %s\n", errb.Error())
+		}
+	}
+
 	v := venom.New()
 	v.RegisterExecutor(exec.Name, exec.New())
 	v.RegisterExecutor(http.Name, http.New())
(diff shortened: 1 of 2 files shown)


contrib/grpcplugins/action/plugin-venom/main.go
```diff
@@ -153,11 +154,17 @@ func (actPlugin *venomActionPlugin) Run(ctx context.Context, q *actionplugin.Act
 		}
 	}
 
+    stopOnFailure, err := strconv.ParseBool(stopOnFailureStr)
```
yesnault, line 157: you have to check if stopOnFailureStr is not empty. 

If empty, stopOnFailureStr should be set to false with no error returned (without that, all existing users using plugin-venom without the stopOnFailureStr set will have an error as: strconv.ParseBool: parsing "": invalid syntax)

yesnault: + I'm not sure that you have gofmt your code

contrib/grpcplugins/action/plugin-venom/main.go
```diff
@@ -76,6 +76,15 @@ func (actPlugin *venomActionPlugin) Run(ctx context.Context, q *actionplugin.Act
 		}, nil
 	}
 
+	stopOnFailure := false
+	if stopOnFailureStr != "" {
+		var errb error
+		stopOnFailure, errb = strconv.ParseBool(stopOnFailureStr)
+		if err != nil {
+			return actionplugin.Fail("Error parsing stopOnFailure value : %s\n", errb.Error())
+		}
```
yesnault, line 85:
```suggestion
		stopOnFailure, err = strconv.ParseBool(stopOnFailureStr)
		if err != nil {
			return actionplugin.Fail("Error parsing stopOnFailure value : %v\n", err)
		}
```

PR #5391, opened 2020-08-23
Fix "cdsctl verify" cmd: Do not use advertised API URL

phsym:
1. Description
cdsctl verify now uses the api url provided in arguments instead of url advertised by CDS verify response. It will display a warning when provided and advertised URLs differ from each other.
This is useful when cdsctl verify is used where the advertised user facing URL is not reachable.
It also makes verify command consistent with login command behavior

1. Related issues
1. About tests
1. Mentions

@ovh/cds

Signed-off-by: phsym <pierre-henri.symoneaux@nokia.com>

### DIFF
--- a/cli/cdsctl/signup.go
+++ b/cli/cdsctl/signup.go
@@ -166,5 +166,8 @@ func signupVerifyFunc(v cli.Values) error {
 		return err
 	}
 
-	return doAfterLogin(client, v, signupresponse.APIURL, sdk.ConsumerLocal, signupresponse)
+	if apiURL != signupresponse.APIURL {
+		fmt.Println("WARNING: The advertised API URL differs from the provided URL")
+	}
+	return doAfterLogin(client, v, apiURL, sdk.ConsumerLocal, signupresponse)
 }


richardlt:
Hi, thanks for the PR. I'm trying to understand in which case your command line is able to login but not verify, the url that is given in the verify URL is not the same as the one you provided to login ?

phsym:
Hi,
In fact, the URL used when running signup and login is the same used when running verify. But it's different from the url configured in CDS.

In that case, the verify command fails because it calls the `doAfterLogin()` function with `signupresponse.APIURL` which is the URL sent by the CDS server in response to the call to `client.AuthConsumerLocalSignupVerify()`. My proposal is to use the url given in the command parameters instead of the one sent by the server.

However, when the verify commands fails because of this, verification is still successful, and we can login after. But this is not user friendly.

The login command dosen't call `doAfterLogin()` with `res.APIURL` but instead uses the url provided in arguments

richardlt (approved):

yesnault (approved):

fsamin (approved):

PR #5491, opened 2020-10-09
fix(sdk): workflow as code parser raise error if 2 roots are detected

fsamin:


sdk/exportentities/v2/workflow.go
```diff
@@ -538,6 +538,10 @@ func (e *NodeEntry) processNodeAncestors(name string, w *sdk.Workflow) (bool, er
 
 	switch len(ancestors) {
 	case 0:
+		// If there is already a root node, it is impossible have aonther one
```
yesnault, line 541:
```suggestion
		// If there is already a root node, it is impossible have another one
```

yesnault (approved):

sguiheux (approved):

richardlt (approved):

PR #5547, opened 2020-11-13
refactor(cdn): using 1 redis queue/job

sguiheux:
1. Description
1. Related issues
1. About tests
1. Mentions

@ovh/cds


engine/cdn/cdn_log_store.go
```diff
@@ -14,54 +14,108 @@ import (
...
+		for _, hm := range hms {
+			now := time.Now()
+			l := sdk.Log{
+				JobID:        hm.Signature.JobID,
+				NodeRunID:    hm.Signature.NodeRunID,
+				LastModified: &now,
+				StepOrder:    hm.Signature.Worker.StepOrder,
+				Val:          hm.Msg.Full,
+			}
+			if err := s.Client.QueueSendLogs(ctx, hm.Signature.JobID, l); err != nil {
```
richardlt, line 65: Be careful with this, we should not call cds api for each line. But one time with all the lines.

richardlt (changes requested):

yesnault (approved):

engine/cdn/cdn_log_store.go
sguiheux: Here, we are in a loop that do 1 call by step, so it's OK
Line38->53: aggregation to build 1hm by step
LIne 56 -> 65: 1 call by step

richardlt: ok

richardlt (approved):

fsamin (approved):

PR #5641, opened 2021-01-18
feat(api): add 'sudo' field in logs

fsamin:


sguiheux (approved):

yesnault (approved):

engine/api/services.go
```diff
@@ -36,7 +36,10 @@ func (api *API) getServiceHandler() service.Handler {
 		// Try to load from DB
 		var srvs []sdk.Service
 		var err error
-		if isAdmin(ctx) || isMaintainer(ctx) {
+		if isAdmin(ctx) {
```
richardlt, line 39: Maybe we can only check isMaintainer and remove tracking ?

fsamin: you are right, PU updated

richardlt (approved):

PR #6051, opened 2022-01-06
feat: delete old action Serve Static File

yesnault:


sguiheux (approved):

engine/sql/api/234_drop_static_files.sql
```diff
@@ -0,0 +1,20 @@
+-- +migrate Up
```
fsamin, line 1: shouldn't we rename the table to be able to restore it if rollback ?

fsamin (approved):

richardlt (approved):

PR #6095, opened 2022-02-18
feat(contrib): support for alternative in plugin-arsenal

Inozuma:
1. Description
1. Related issues
1. About tests
1. Mentions

@ovh/cds

### DIFF
--- a/contrib/integrations/arsenal/plugin-arsenal/main.go
+++ b/contrib/integrations/arsenal/plugin-arsenal/main.go
@@ -123,7 +123,6 @@ func (e *arsenalDeploymentPlugin) Run(ctx context.Context, q *integrationplugin.
 	var (
 		application       = getStringOption(q, "cds.application")
 		workflowRunNumber = getStringOption(q, "cds.run.number")
-		workflowNodeID    = getStringOption(q, "cds.node.id")
 		arsenalHost       = getStringOption(q, "cds.integration.deployment.host")
 		deploymentToken   = getStringOption(q, "cds.integration.deployment.deployment.token", "cds.integration.deployment.token")
 		alternative       = getStringOption(q, "cds.integration.deployment.alternative.config")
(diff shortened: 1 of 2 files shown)


contrib/integrations/arsenal/plugin-arsenal/main.go
```diff
@@ -73,145 +108,293 @@ const deployData = `{
...
+				return fail("failed to unmarshal alternative config: %v", err)
+			}
 
-	// Do the request
-	res, err := httpClient.Do(req)
-	if err != nil {
-		return fail("Error: Post %s/deploy failed: %v. Please check you integration configuration", arsenalHost, err)
+			// Create alternative on /alternative
+			fmt.Printf("Creating alternative %s on Arsenal...\n", altConfig.Name)
+			if err = arsenalClient.upsertAlternative(altConfig); err != nil {
```
fsamin, line 170: I guess somewhere we have to set the workflow_run_id associated with the alternative to be able to delete it later

fsamin, line 170: For readability it could be nice to print the interpolated alternative config (json prettyfied)

fsamin (changes requested):

contrib/integrations/arsenal/plugin-arsenal/main.go
Inozuma: Indeed, will add it in alternative's options.

Inozuma: It is actually printed, but only in case of failure. Do you think it should always be printed?

Inozuma: Added `cds.run.number`, `cds.node.id` and `cds.application` in `Options`.

Inozuma: Added a print to a prettified json with enriched cds metadata.

sguiheux (approved):

yesnault (approved):

contrib/integrations/arsenal/plugin-arsenal/main.go
Inozuma: Alright, removed

fsamin (approved):

richardlt (approved):

PR #6429, opened 2023-01-24
feat(api,cli,ui) disable project key

fsamin:


richardlt (approved):

engine/sql/api/263_key_disabled.sql
```diff
@@ -0,0 +1,7 @@
+
+-- +migrate Up
+ALTER TABLE project_key ADD COLUMN disabled BOOLEAN NOT NULL DEFAULT FALSE;
+UPDATE project_key SET disabled = false;
+
+-- +migrate Down
+ALTER TABLE project_key DROP COLUMN BOOLEAN;
```
sguiheux, line 7:
```suggestion
ALTER TABLE project_key DROP COLUMN disabled;
```

engine/sql/api/263_key_disabled.sql
```diff
@@ -0,0 +1,7 @@
+
+-- +migrate Up
+ALTER TABLE project_key ADD COLUMN disabled BOOLEAN NOT NULL DEFAULT FALSE;
+UPDATE project_key SET disabled = false;
```
sguiheux, line 4: DEFAULT false is not updating all existing rows ?

If there is lots of values the right pattern is:

ALTER TABLE project_key ADD COLUMN disabled BOOLEAN NOT NULL;
UPDATE project_key SET disabled = false;
ALTER TABLE "project_key" ALTER COLUMN disabled SET DEFAULT false;

sguiheux (changes requested):

engine/sql/api/263_key_disabled.sql
fsamin: > DEFAULT false is not updating all existing rows ?

nope

there are not so much rows (1468) in the table

sguiheux (approved):

PR #6438, opened 2023-02-02
Add uuidv4 helper function

sixstone-qq:
1. Description

Using ovh/venom, I found the variables helpers are not using latest version
from `Masterminds/sprig` package so I backported the one that could be really
useful (`uuidv4` generation) here.

Taken from
https://github.com/Masterminds/sprig/blob/3ac42c7bc5e4be6aa534e036fb19dde4a996da2e/functions.go#L355

1. Related issues

I didn't use `sdk.UUID()` because it's a different package and I see there is a
different Go module.

Let me know what you think.

1. About tests
1. Mentions

@ovh/cds

### DIFF
--- a/sdk/interpolate/go.mod
+++ b/sdk/interpolate/go.mod
@@ -4,6 +4,7 @@ go 1.19
 
 require (
 	github.com/aokoli/goutils v1.1.0
+	github.com/google/uuid v1.3.0
 	github.com/huandu/xstrings v1.2.0
 	github.com/spf13/cast v1.4.1
 	github.com/stretchr/testify v1.5.1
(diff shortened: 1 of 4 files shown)


richardlt:
Should be possible to reuse sdk.UUID() rather than importing a new lib.

richardlt (changes requested):
Should be possible to reuse sdk.UUID() rather than importing a new lib.

yesnault:
Hi, you can define helper into venom only, not needed to add it into cds.

sixstone-qq:
> Hi, you can define helper into venom only, not needed to add it into cds.

Thanks a lot for your response, I'd like to include all new functions from https://github.com/Masterminds/sprig but this is not currently possible to only do that in venom because `template.FuncMap` is being wrapper with [wrapHelpers](https://github.com/ovh/cds/blob/master/sdk/interpolate/interpolate_helper.go#L86) function which is private so I can't override `InterpolateHelperFuncs` function. Which action path do you suggest me to follow?

Thanks a lot!

PR #6504, opened 2023-03-11
fix: doc generate error handling

alrs:
This fixes a dropped error in `sdk/doc`.

@ovh/cds


alrs:
I'll resubmit this without the stray `engine/worker/internal/input` business.

PR #7073, opened 2024-08-08
feat(contrib): add run result plugin

sguiheux:
1. Description
1. Related issues
1. About tests
1. Mentions

@ovh/cds


tests/08_v2_matrix.yml
sguiheux, line 1: Removed because we cannot use fromJSON function in venom test

richardlt (approved):

yesnault (approved):

Issue #7213, opened 2024-11-08
Add fromFile interpolate helper 


LuBashQ:
Hi, as I have pointed in https://github.com/ovh/venom/issues/822, it would be nice to have an interpolate helper that would read a file (either relative of absolute path) and return the contents of said file.

Doing so would open to cleaner assertions in `venom`, as the comparison values can then be stored in separate files and not added as variables. At least for the use cases I am facing in my day job, I prefer to store and have syntax highlighting and clearly stated changes/commits on said files, and those files could be shared as-is between several utilities.

PR #7214, opened 2024-11-09
feat(interpolation): add fromFile interpolation helper

LuBashQ:
1. Description

Add interpolation helper that reads a file path and returns the contents of the file if exists, empty string otherwise. The method is intended to be used for example chained to `toJSON` helper, given a correctly formatted JSON file.

2. Related issues

#7213

3. About tests

Added unit tests with both existing and non-existing paths.

### DIFF
--- a/sdk/interpolate/interpolate_helper.go
+++ b/sdk/interpolate/interpolate_helper.go
@@ -9,6 +9,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/url"
+	"os"
 	"path"
 	"reflect"
 	"strconv"
@@ -89,6 +90,7 @@ func init() {
 		"urlencode": func(s string) string { return url.QueryEscape(s) },
 		"dirname":   func(s string) string { return path.Dir(s) },
 		"basename":  func(s string) string { return path.Base(s) },
+		"fromFile":  fromFile,
 	})
 }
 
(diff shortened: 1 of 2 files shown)


richardlt:
Hi, I understand your request about Venom but as this interpolation is also used by CDS api we can't safely approve the PR as we don't want this interpolate to read files in that context.
Venom should use its own set of helpers for that purpose, you can submit a PR on Venom repository if needed.

LuBashQ:
The reason why the PR was opened here is that venom uses cds' `interpolate` package to resolve helper functions https://github.com/ovh/venom/blob/f3548d1573594480413f728e4a57f0f5130b36e8/types_executor.go#L275, which are filled in here https://github.com/ovh/cds/blob/fafc78f7d74c8dad863e34d28cdef8ba7b4cdfa2/sdk/interpolate/interpolate.go#L166.

My question now is? Would it be then a good option to treat the return from `interpolate.Do()` in Venom as a template and then further process it with the venom-specific helper functions? Correct me if I'm wrong, but from what I understand `interpolate.Do()` in case of unknown helpers, it will just keep them as-is https://github.com/ovh/cds/blob/fafc78f7d74c8dad863e34d28cdef8ba7b4cdfa2/sdk/interpolate/interpolate.go#L153-L160, so could it then be processed further by venom?

PR #7328, opened 2025-03-02
fix: fix the wrong error return value

cangqiaoyuzhuo:
In fact, err is incorrect and should be lastErr.

@ovh/cds


yesnault (approved):

richardlt:
Fixed by #7354 https://github.com/ovh/cds/pull/7354

PR #7432, opened 2025-06-04
feat(cds v2): add coalesce and default functions for ${{ ... }} templating

kilianpaquier:
For simpler workflows and jobs, I would like to add `default` and `coalesce` functions to templating functions to avoid having shell scripts steps to fallback on empty values (mostly string, but I think it's OK to handle any type of input).

@ovh/cds

### DIFF
--- a/sdk/action_parser_funcs.go
+++ b/sdk/action_parser_funcs.go
@@ -53,12 +53,40 @@ var (
 		"match":        match,
 		"replace":      replace,
 		"contextValue": contextValue,
+		"default":      dfault,
+		"coalesce":     coalesce,
 	}
 )
 
 type ActionFunc func(ctx context.Context, a *ActionParser, inputs ...interface{}) (interface{}, error)
 
-func replace(_ context.Context, a *ActionParser, inputs ...interface{}) (interface{}, error) {
+func dfault(_ context.Context, a *ActionParser, inputs ...interface{}) (interface{}, error) {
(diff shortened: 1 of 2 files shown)


fsamin (approved):

sguiheux:
Waiting PR https://github.com/ovh/cds/pull/7437 to be able to run CDS from fork

sdk/action_parser_funcs.go
```diff
@@ -53,12 +53,41 @@ var (
...
-func replace(_ context.Context, a *ActionParser, inputs ...interface{}) (interface{}, error) {
+func dfault(_ context.Context, a *ActionParser, inputs ...interface{}) (interface{}, error) {
+	switch len(inputs) {
+	case 0:
+		return nil, nil // default
+	case 1:
+		return inputs[0], nil // "var" | default OR default "var"
+	case 2:
+		// second arg is the first input value since in case of | (pipe) it's added as last argument
+		if empty(inputs[1]) {
```
sguiheux, line 71: usage of the function must be.  ${{ default(input0, input1) }} there is no pipe.
If input0 is not empty, you should return input0

sguiheux (changes requested):

sdk/action_parser_funcs.go
kilianpaquier: Done 🙂

PR #7460, opened 2025-07-16
fix(api): delete artifacts in parallel

sguiheux:
1. Description
1. Related issues
1. About tests
1. Mentions

@ovh/cds

### DIFF
--- a/engine/api/purge/purge_v2.go
+++ b/engine/api/purge/purge_v2.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/go-gorp/gorp"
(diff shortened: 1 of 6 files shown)


engine/api/purge/purge_v2.go
```diff
@@ -363,29 +367,40 @@ func DeleteArtifactsFromRepositoryManagerV2(ctx context.Context, db gorp.SqlExec
 	props.AddProperty("ovh.to_delete", "true")
 	props.AddProperty("ovh.to_delete_timestamp", strconv.FormatInt(time.Now().Unix(), 10))
 
+	results := make(chan bool, len(runResults))
```
richardlt, line 370: Maybe using a wait group here is easier to maintain.

richardlt (approved):

engine/api/purge/purge_v2.go
sguiheux: fixed

PR #7485, opened 2025-08-19
feat(api): v2 template add 'name' param

richardlt:
1. Description
1. Related issues
1. About tests
1. Mentions

@ovh/cds


JayBeeDe:
`${{cds.workflow}}` requires the workflow to run, which is not the case when initializing the workflow from the template

sguiheux:
So maybe the example in the test case is not revelant.
If it's to customize value as variableset name or things like this it's ok

PR #7696, opened 2026-03-27
fix(api): when create or update an item, return it not the variableset

sguiheux:
1. Description
1. Related issues
1. About tests
1. Mentions

@ovh/cds


richardlt (approved):

yesnault (approved):

richardlt:
Could be nice to reload the data from database without clear values rather than retuning the original request with some modifications.

PR #7717, opened 2026-04-14
fix: Add retry on http status code 422 (Unprocessable Entity)

ncharaf:
1. Description
Add retry on http status code 422 (Unprocessable Entity) when deploying on Arsenal
1. Related issues
The deploy fail if we got an http status 422 during deployment

@ovh/cds


fsamin:
Retrying a 422 error doesn't make sens. It will always return 422.

PR #7782, opened 2026-07-02
feat(contrib): support glob patterns in addRunResult path

fsamin:


hiSandog:
For glob support in `addRunResult`, the edge cases are probably more important than the happy path. Tests should cover no matches, multiple matches with stable ordering, escaped glob characters in file names, and whether paths outside the workspace are rejected or normalized safely.

yesnault (approved):

sguiheux (approved):

PR #7837, opened 2026-09-03
perf(cache): walk the keyspace with SCAN instead of blocking on KEYS

### LINKED ISSUE #7833: Redis KEYS in a hot loop blocks the whole instance for every service
## What happens

`RedisStore.Keys()` in `engine/cache/redis.go` issues the Redis `KEYS` command.

`KEYS` is O(N) over the entire keyspace, and Redis executes commands on a single
thread — `io-threads` parallelises socket reads and protocol parsing only, not
command execution. So for as long as a `KEYS` call runs, every other client of
that Redis waits, whatever it was asking for.

The CDN calls this in a hot loop. `engine/cdn/cdn_log_engine.go`:

```go
func (s *Service) waitingJobs(ctx context.Context) {
	for {
		time.Sleep(250 * time.Millisecond)
		...
		keyListQueue := cache.Key(keyJobLogQueue, "*")
		listKeys, err := s.Cache.Keys(keyListQueue)
```

Four full keyspace scans per second, forever, for the lifetime of the service.

## Measurements

From a CDS instance with a 143k-key Redis (`redis_version 7.4.3`, `io-threads 1`):

```
SLOWLOG LEN                     128
entries matching KEYS cdn:log:job:*   128   ← every single entry
per-call duration                27,085–35,560 µs  (27–35 ms)
DBSIZE                           143,523
instantaneous_ops_per_sec        252
redis CPU under load             88% of a core
```

Every entry in the slow-command log was this one call. Nothing else in the
system was slow enough to register.

The knock-on effect is not confined to the CDN — every CDS service shares this
Redis, so the stall lands on lock acquisition, queue pops and cache reads
belonging to unrelated components. On our instance a repository analysis sat
unclaimed for **136 seconds** while its poller ticked every 5 seconds, because
each tick's Redis work was queued behind these scans.

## Severity depends on keyspace size

The blocking is inherent to `KEYS` on any Redis deployment, but the cost is
O(keyspace). A small installation may never notice; ours became unusable. Worth
saying plainly so the report isn't read as universal breakage.

## Proposed fix

Walk the keyspace with `SCAN` instead. Same result, never blocks for more than
one batch. `SCAN` can return duplicates across cursor steps (de-duplicated) and
is a moving view rather than a snapshot — for a caller polling to find work to
pick up, that is what `KEYS` gave it in practice anyway.

PR: https://github.com/ovh/cds/pull/7837

### After the change, same instance

```
SLOWLOG LEN                     0        (was 128, all KEYS)
command latency                 0.00 ms  (was 27–35 ms blocking)
instantaneous_ops_per_sec       153      (was 252)
redis CPU                       47.6%    (was 88%)
CDN "unable to list jobs queues" errors   0
```

Batch size matters: at `COUNT 1000` the 143k keyspace takes ~140 cursor steps
and the command rate went *up* (252 → 523/s) before settling at ~15 steps and
153/s with `COUNT 10000`.

gsalingu:
1. Description

`RedisStore.Keys()` used the Redis `KEYS` command. `KEYS` is O(N) over the whole
keyspace and Redis executes commands on a single thread, so each call blocks
every other client of that Redis for its duration. The CDN's `waitingJobs` loop
calls it four times a second for the lifetime of the service.

On a 143k-key instance every one of the 128 entries in `SLOWLOG` was this call,
at 27–35 ms each. Because all CDS services share the Redis, the stall landed on
locks, queue pops and cache reads belonging to unrelated components.

This walks the keyspace with `SCAN` instead. `SCAN` returns the same keys and
never blocks for more than one batch. It can repeat keys across cursor steps, so
they are de-duplicated, and it is a moving view rather than a snapshot — for a
caller polling to find work to pick up, that is what `KEYS` provided in practice.

Batch size is a constant rather than configuration: nothing outside Redis
constrains it, a wrong value costs efficiency rather than correctness, and there
is no state an operator can see that the code cannot. It does need to be large —
at 1000 the same keyspace takes ~140 cursor steps and the command rate rose from
252/s to 523/s; at 10000 it is ~15 steps and 153/s.

Measured on the same instance, before and after:

| | KEYS | SCAN (10000) |
|---|---|---|
| SLOWLOG entries | 128, all this call | 0 |
| command latency | 27–35 ms blocking | 0.00 ms |
| ops/sec | 252 | 153 |
| redis CPU | 88% | 47.6% |
| CDN errors | — | 0 |

1. Related issues

Closes #7833

1. About tests

Two tests in `engine/cache/redis_test.go`, against a live Redis:

- `TestKeysWalksWholeKeyspace` — 250 keys under a random prefix plus one key
  outside it. The walk is driven with a batch of 10 so a small keyspace still
  forces roughly twenty-five cursor steps and the loop itself is under test
  rather than a single round trip. Asserts the result is complete, contains no
  key twice (SCAN may repeat across steps), and excludes the non-matching key.
  It then repeats through the exported `Keys()` on the production batch and
  asserts the same set, pinning the two batch sizes to agree.
- `TestKeysReturnsEmptyForUnmatchedPattern` — a pattern matching nothing must
  terminate and return empty rather than error or spin on a non-zero cursor.

To make the cursor loop reachable with a small keyspace the walk is split into
an unexported `scanKeys(pattern, batch)` that `Keys()` calls with
`keysScanBatch`. If you would rather not carry that seam, say so and I will
drive the test through `Keys()` alone with a keyspace larger than the batch.

These pin behaviour rather than catch a regression, and deliberately so: `KEYS`
returned correct results too, and the defect being fixed is latency, not
correctness. The correctness risk lives entirely in the replacement, which is
what they cover; the performance claim is the production measurement above.

The full `engine/cache` package suite passes against a live Redis.

1. Mentions

@ovh/cds

### DIFF
--- a/engine/cache/redis.go
+++ b/engine/cache/redis.go
@@ -105,15 +105,73 @@ func (s *RedisStore) Ping() error {
 	return nil
 }
 
+// keysScanBatch is how many keys one SCAN cursor step asks for.
+//
+// Not configurable on purpose. Nothing outside Redis constrains it, a wrong
+// value costs efficiency rather than correctness, and there is no state an
+// operator can see that the code cannot — so a knob here would only ask
+// someone to guess at a number that has one right answer.
+//
+// It does need to be large. At 1000 a keyspace of a hundred and forty thousand
+// keys takes about a hundred and forty cursor steps, and the CDN asks four
+// times a second, which doubled this instance's command rate from 250/s to
+// 520/s the moment SCAN replaced KEYS. At 10000 the same walk is about
+// fifteen steps. Redis handles a single SCAN step of this size in well under
+// a millisecond, so the larger batch costs nothing that the round trips did
+// not cost more of.
+const keysScanBatch = 10000
+
+// Keys returns the keys matching a pattern.
+//
+// This walks the keyspace with SCAN rather than asking for KEYS. KEYS is O(N)
+// over the whole keyspace and, because Redis executes commands on one thread,
+// it blocks every other client for as long as it runs. The CDN's waitingJobs
+// loop calls this four times a second forever; against a keyspace of a hundred
+// and forty thousand keys each call measured 27-35ms in the slowlog, so that
+// loop alone held the server for roughly a tenth of all available time and
+// added a stall to every other command — locks, queue pops, cache reads — for
+// every service sharing this Redis. It is why a repository analysis could sit
+// unclaimed for over two minutes while the poller ticked every five seconds.
+//
+// SCAN returns the same keys without ever blocking for more than one batch.
+// It may return duplicates across cursor steps, which the set here removes,
+// and it is a moving view of a live keyspace rather than a snapshot — for a
+// caller that is polling for work to pick up, that is the same guarantee KEYS
+// gave it in practice.
 func (s *RedisStore) Keys(pattern string) ([]string, error) {
 	if s.Client == nil {
 		return nil, sdk.WithStack(fmt.Errorf("redis> cannot get redis client"))
 	}
-	keys, err := s.Client.Keys(context.Background(), pattern).Result()
-	if err != nil {
-		return nil, sdk.WrapError(err, "redis> cannot list keys: %s", pattern)
+	return s.scanKeys(pattern, keysScanBatch)
+}
+
+// scanKeys walks the keyspace one cursor step at a time.
+//
+// Split out from Keys so the multi-step path can be exercised with a batch
+// small enough to force several cursor steps over a small keyspace; production
+// callers always go through Keys.
+func (s *RedisStore) scanKeys(pattern string, batch int64) ([]string, error) {
+	ctx := context.Background()
+	seen := make(map[string]struct{})
+	keys := make([]string, 0)
+	var cursor uint64
+	for {
+		found, next, err := s.Client.Scan(ctx, cursor, pattern, batch).Result()
+		if err != nil {
+			return nil, sdk.WrapError(err, "redis> cannot list keys: %s", pattern)
+		}
+		for _, k := range found {
+			if _, dup := seen[k]; dup {
+				continue
+			}
+			seen[k] = struct{}{}
+			keys = append(keys, k)
+		}
+		cursor = next
+		if cursor == 0 {
+			return keys, nil
+		}
 	}
-	return keys, nil
 }
 
 // Get a key from redis
--- a/engine/cache/redis.go
+++ b/engine/cache/redis.go
@@ -108,18 +108,11 @@ func (s *RedisStore) Ping() error {
 // keysScanBatch is how many keys one SCAN cursor step asks for.
 const keysScanBatch = 10000
 
-// keysScanTimeout bounds a whole keyspace walk.
-//
-// KEYS was one round trip; a walk is several, each with the client's own
-// retries behind it, so a degraded Redis could hang a caller for much longer
-// than before. The walk is a handful of steps of well under a millisecond of
-// server time each, so this is several orders of magnitude of headroom and
-// only ever trips on a Redis that is not answering.
+// keysScanTimeout bounds a whole walk, however many cursor steps it takes, so
+// a Redis that stops answering cannot hold a caller indefinitely.
 const keysScanTimeout = 30 * time.Second
 
 // Keys returns the keys matching a pattern, walking the keyspace with SCAN.
-// KEYS is O(N) over the whole keyspace and blocks every other client of the
-// instance for as long as it runs.
 func (s *RedisStore) Keys(pattern string) ([]string, error) {
 	if s.Client == nil {
 		return nil, sdk.WithStack(fmt.Errorf("redis> cannot get redis client"))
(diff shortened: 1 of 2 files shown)


engine/cache/redis.go
```diff
@@ -105,15 +105,73 @@ func (s *RedisStore) Ping() error {
 	return nil
 }
 
+// keysScanBatch is how many keys one SCAN cursor step asks for.
```
sguiheux, line 108: There is no reason to talk about CDN here. Please config your IA to not write its context on comment please
One line comment must be enough

engine/cache/redis.go
```diff
@@ -105,15 +105,73 @@ func (s *RedisStore) Ping() error {
...
+// It does need to be large. At 1000 a keyspace of a hundred and forty thousand
+// keys takes about a hundred and forty cursor steps, and the CDN asks four
+// times a second, which doubled this instance's command rate from 250/s to
+// 520/s the moment SCAN replaced KEYS. At 10000 the same walk is about
+// fifteen steps. Redis handles a single SCAN step of this size in well under
+// a millisecond, so the larger batch costs nothing that the round trips did
+// not cost more of.
+const keysScanBatch = 10000
+
+// Keys returns the keys matching a pattern.
```
sguiheux, line 124: Same as above

engine/cache/redis_test.go
```diff
@@ -109,3 +109,76 @@ func TestDequeueJSONRawMessagesWithContextMaxTimeout(t *testing.T) {
...
+
+// TestKeysWalksWholeKeyspace covers the SCAN-based Keys(). SCAN differs from
+// KEYS in ways that matter to callers: it pages through a cursor, it can return
+// the same key more than once, and it needs several round trips when the
+// keyspace is larger than one batch.
+//
+// The batch is deliberately tiny here so a small keyspace still forces many
+// cursor steps; production uses keysScanBatch. The result must be complete,
+// free of duplicates, and correctly filtered by pattern.
+func TestKeysWalksWholeKeyspace(t *testing.T) {
```
sguiheux, line 121: You talk about duplicates, but your are not generating duplicate entries, could you improve the test?

sguiheux (commented):
1. DeleteAll still uses KEYS

[redis.go:208](vscode-webview://087te31n46236q84ai3j7uqca1nvi33nt9vg83fvsj609la5s1fb/engine/cache/redis.go#L208) is untouched. Is there a reason of that ?


2. No cancellation, and the failure window grows

The loop is now up to 15 round trips on context.Background(), with MaxRetries: 10 on the client. On a degraded Redis a single Keys call can hang much longer than before. The Store interface not taking a context is pre-existing, but this PR makes it worse. 

Not blocking, but worth noting.


On your installation, do you use the same redis for api/cdn/repositories/vcs/hooks ?

sguiheux:
@richardlt @fsamin @yesnault can we have you pov on this PR ?

sguiheux (changes requested):

gsalingu:
All four points, three commits, no force push.

**1. `DeleteAll` still used KEYS.** No reason beyond scope: `Keys` was the call in the hot loop and the one I measured. The blocking is a property of the command rather than of the caller, so `DeleteAll` walks the same cursor now. It has a test of its own over a keyspace larger than one batch, since a truncated walk would silently leave keys behind where the single-shot listing did not.

**2. No cancellation, and the failure window grows.** Fair, and it was a real regression: KEYS was one round trip, the walk is several, each with `MaxRetries` behind it. The walk now has a single 30s budget for the whole cursor loop. A walk is a handful of steps of well under a millisecond of server time each, so that is several orders of magnitude of headroom and can only trip on a Redis that is not answering — and a degraded Redis now costs a caller no more than the one KEYS call did. The `Store` interface still takes no context, which is pre-existing; this stops the change making it worse.

**3. Comments.** Cut to one line each.

**4. Duplicates.** You are right, the assertion was vacuous. A live Redis only returns a key twice if it happens to resize its hash table mid-walk, so nothing in that test produced a duplicate and it passed whether or not the code de-duplicated at all. The de-duplication is now `appendUnseen`, driven directly with the input a resize produces: a key repeated on a later cursor step, and a key repeated within one step. `TestKeysWalksWholeKeyspace` keeps what it can genuinely assert against a real Redis — complete across cursor pages, still filtered by pattern.

**Same Redis for api/cdn/repositories/vcs/hooks?** Yes, one instance shared by all five. That is why the stall was not confined to the CDN: `redis_version 7.4.3`, `io-threads 1`, 143k keys, and every entry in `SLOWLOG` was this call.

engine/cache/redis.go
gsalingu: On it, sorry about this!

gsalingu: Done in b798782 — one line, no CDN and no incident context. 8500181 then went over every comment in this PR the same way.

engine/cache/redis.go
gsalingu: Same, in b798782.

engine/cache/redis_test.go
gsalingu: You were right, the assertion was vacuous: a live Redis repeats a key only if it happens to resize its hash table mid-walk, so nothing in that test produced one and it passed whether or not the code de-duplicated.

fb3fd7f8 moves the de-duplication into `appendUnseen` and drives it with the input a resize produces — a key repeated on a later cursor step, and a key repeated within one step. `TestKeysWalksWholeKeyspace` keeps only what it can genuinely assert against a real Redis, and `DeleteAll` gets a test of its own now that it walks the same cursor.

engine/cache/redis.go
gsalingu: I am really sorry about the extra mental load here... 
I am working on my setup to fix this specifically

Overall I do hope I am helping with these PR's, I am *REALLY*  thankful for the tool.
I've tried a bunch of CI tools, so far CDS is really the best out there!

PR #7838, opened 2026-09-03
perf(api): read entity files concurrently during a repository analysis

### LINKED ISSUE #7834: Repository analysis reads entity files serially, so its duration scales with entity count
## What happens

`getCdsFilesOnVCSDirectory` in `engine/api/v2_repository_analyze.go` reads each
entity file with its own `client.GetContent` call, one after another:

```go
for _, c := range contents {
	if c.IsFile && (strings.HasSuffix(c.Name, ".yml") || ...) {
		vcsContent, err := client.GetContent(ctx, repoName, commit, filePath)
```

Every read is a network round trip to the VCS service and on to the forge, so a
repository analysis costs one serial round trip per entity. Analysis duration
therefore scales with **how many entities a repository declares**, not with the
work involved in analysing them.

Gitea/GitHub/GitLab/Forgejo take this path. Bitbucket takes
`getCdsArchiveFileOnRepo`, which fetches once.

## Measurements

Two repositories on the same instance, same commit-to-analysis path:

| repository | entities | analysis |
|---|---|---|
| ~130 `.cds` entities | ~130 | **5 min 30 s** |
| few entities | few | 36 s |

The VCS service logged **2,453 requests in 15 minutes**, almost all single-file
reads serving one analysis. Per-request latency to the forge on this instance:

```
n=637   p50=1237ms   p90=2295ms   max=4422ms
```

130 files × ~1.2 s ≈ 160 s of pure waiting, in series, per analysis.

## Proposed fix

Split listing from reading. Listing stays sequential — one call per directory,
a handful in total. The reads, which are what scales, are done concurrently with
a bounded worker group.

PR: https://github.com/ovh/cds/pull/7838

Result on the 130-entity repository: **330 s → 114 s**.

## Open question for maintainers

The PR uses a constant of 8. The right ceiling is a property of whatever fronts
the forge rather than of CDS — on our instance that is a single reverse proxy
shared by the whole estate, and we run it as a configuration value
(`entity.fileFetchConcurrency`) for that reason. Happy to submit it as
configuration instead if you would prefer that; the constant keeps the diff
small for review.

gsalingu:
1. Description

`getCdsFilesOnVCSDirectory` read each entity file with its own `GetContent`
call, in series. Every read is a round trip to the VCS service and on to the
forge, so an analysis cost one serial round trip per entity and its duration
scaled with how many entities a repository declares rather than with the work of
analysing them.

On our instance a repository with ~130 `.cds` entities took 5 min 30 s to
analyse while one with few entities took 36 s, and the VCS service logged 2,453
requests in fifteen minutes — nearly all single-file reads serving one analysis.
Round-trip latency to the forge was p50 1237 ms, so ~130 files in series is
around 160 s of pure waiting.

The listing is unchanged and still sequential: it is one call per directory, a
handful in total. Only the reads, which are what scales, are fanned out, with a
bounded group so the forge is not flooded.

Result on the 130-entity repository: 330 s → 114 s.

Note this path is taken by Gitea, GitHub, GitLab and Forgejo; Bitbucket already
fetches an archive in one call via `getCdsArchiveFileOnRepo`.

One note on the bound. It is a constant here, at 8. We run this as a
configurable value set to 16 and saw bursts of 14-15 concurrent completions in
the VCS service logs, so the ceiling does bind in practice rather than being
theoretical. A constant seemed right for upstream, since the sensible value
depends on what fronts the forge rather than on anything an operator can
readily see — but if you would rather it were configurable, that is a small
change and I am happy to make it.

1. Related issues

Closes #7834

1. About tests

Two tests in `engine/api/v2_repository_analyze_test.go`, against a live Postgres
with the VCS service mocked:

- `TestGetCdsFilesOnVCSDirectoryReadsEveryFileConcurrentlyAndBounded` — a tree
  of twenty-six entity files across a directory and a subdirectory, plus a
  `.txt` that must be ignored. Asserts the same result the serial version
  produced: every entity file exactly once, keyed by full path, base64-decoded,
  and nothing that is not an entity file. It also holds each mocked read open
  briefly and tracks the peak number in flight, asserting that peak exceeds one
  — serial code can never do that — and never exceeds the bound.
- `TestGetCdsFilesOnVCSDirectoryPropagatesReadFailures` — a failing read must
  fail the whole call. Worth pinning under concurrency: a dropped error would
  turn a partial read into an analysis that silently sees fewer entities than
  the repository declares.

too).

1. Mentions

@ovh/cds

### DIFF
--- a/engine/api/v2_repository_analyze.go
+++ b/engine/api/v2_repository_analyze.go
@@ -15,6 +15,9 @@ import (
 	"regexp"
 	"sort"
 	"strings"
+	"sync"
+
+	"golang.org/x/sync/errgroup"
 	"time"
 
 	"github.com/go-gorp/gorp"
@@ -2009,6 +2012,48 @@ func (api *API) analyzeCommitSignatureThroughOperation(ctx context.Context, anal
 	return keyId, analyzeError, nil
 }
 
+// cdsFileFetchConcurrency bounds how many entity files are read from the VCS
+// at once.
+const cdsFileFetchConcurrency = 8
+
+// listCdsFilePaths walks the .cds tree and returns the entity files in it.
+// Listing is one call per directory, a handful in total; reading the files is
+// the part that scales with the repository, so it is done separately and can
+// then be done in parallel.
+func (api *API) listCdsFilePaths(ctx context.Context, client sdk.VCSAuthorizedClientService, repoName, commit, directory string) ([]string, error) {
+	contents, err := client.ListContent(ctx, repoName, commit, directory, "0", "100")
+	if err != nil {
+		return nil, sdk.WrapError(err, "unable to list content on commit [%s] in directory %s: %v", commit, directory, err)
+	}
+	paths := make([]string, 0, len(contents))
+	for _, c := range contents {
+		if c.IsFile && (strings.HasSuffix(c.Name, ".yml") || strings.HasSuffix(c.Name, ".yaml")) {
+			paths = append(paths, directory+"/"+c.Name)
+		}
+		if c.IsDirectory {
+			sub, err := api.listCdsFilePaths(ctx, client, repoName, commit, directory+"/"+c.Name)
+			if err != nil {
+				return nil, err
+			}
+			paths = append(paths, sub...)
+		}
+	}
+	return paths, nil
+}
+
+// getCdsFilesOnVCSDirectory reads every entity file under a directory.
+//
+// The files are read concurrently. Read one after another, an analysis costs
+// one network round trip per entity in series, so its duration grows with the
+// number of entities a repository declares rather than with the work involved:
+// a repository with about a hundred and thirty entities took five and a half
+// minutes to analyse while one with far fewer took thirty-six seconds, and the
+// VCS service logged two and a half thousand requests in fifteen minutes,
+// nearly all single-file reads for one analysis. None of that is compute, so
+// it parallelises directly.
+//
+// Bounded rather than unbounded: every read crosses whatever fronts the forge,
+// and that hop, not CDS, is what sets the safe ceiling.
 func (api *API) getCdsFilesOnVCSDirectory(ctx context.Context, analysis *sdk.ProjectRepositoryAnalysis, vcsName, repoName, commit, directory string) (map[string][]byte, error) {
 	ctx, next := telemetry.Span(ctx, "api.getCdsFilesOnVCSDirectory")
 	defer next()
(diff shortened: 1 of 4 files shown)


engine/api/v2_repository_analyze.go
```diff
@@ -2009,6 +2012,48 @@ func (api *API) analyzeCommitSignatureThroughOperation(ctx context.Context, anal
 	return keyId, analyzeError, nil
 }
 
+// cdsFileFetchConcurrency bounds how many entity files are read from the VCS
+// at once.
+const cdsFileFetchConcurrency = 8
```
sguiheux, line 2017: Should be in CDS configuration. Depending your VCS you can have Ratelimit on simultaneous request, call/mn etc.. ( ex: Github ), so it can be useful to be able to modified the value regarding the activity of a CDS instance

engine/api/v2_repository_analyze.go
```diff
@@ -2009,6 +2012,48 @@ func (api *API) analyzeCommitSignatureThroughOperation(ctx context.Context, anal
...
+
+// listCdsFilePaths walks the .cds tree and returns the entity files in it.
+// Listing is one call per directory, a handful in total; reading the files is
+// the part that scales with the repository, so it is done separately and can
+// then be done in parallel.
```
sguiheux, line 2022:
```suggestion
```

engine/api/v2_repository_analyze.go
```diff
@@ -2009,6 +2012,48 @@ func (api *API) analyzeCommitSignatureThroughOperation(ctx context.Context, anal
...
+// one network round trip per entity in series, so its duration grows with the
+// number of entities a repository declares rather than with the work involved:
+// a repository with about a hundred and thirty entities took five and a half
+// minutes to analyse while one with far fewer took thirty-six seconds, and the
+// VCS service logged two and a half thousand requests in fifteen minutes,
+// nearly all single-file reads for one analysis. None of that is compute, so
+// it parallelises directly.
+//
+// Bounded rather than unbounded: every read crosses whatever fronts the forge,
+// and that hop, not CDS, is what sets the safe ceiling.
```
sguiheux, line 2056:
```suggestion
```

engine/api/v2_repository_analyze.go
```diff
@@ -2018,33 +2063,34 @@ func (api *API) getCdsFilesOnVCSDirectory(ctx context.Context, analysis *sdk.Pro
...
+	paths, err := api.listCdsFilePaths(ctx, client, repoName, commit, directory)
 	if err != nil {
-		return nil, sdk.WrapError(err, "unable to list content on commit [%s] in directory %s: %v", commit, directory, err)
+		return nil, err
 	}
-	for _, c := range contents {
-		if c.IsFile && (strings.HasSuffix(c.Name, ".yml") || strings.HasSuffix(c.Name, ".yaml")) {
-			filePath := directory + "/" + c.Name
-			vcsContent, err := client.GetContent(ctx, repoName, commit, filePath)
+
```
sguiheux, line 2070: 1. errgroup spawns bare goroutines with no panic recovery (x/sync deliberately removed it). Before this PR, a panic in the read path was caught by the recover in GoRoutines.Exec; now it kills the API process.

2. Once groupCtx is cancelled by a failing read, this loop keeps calling group.Go for every remaining path;

sguiheux (changes requested):

gsalingu:
All three applied, three commits, no force push.

**Configuration.** `entity.fileFetchConcurrency`, default 8, applied in `Serve` next to the other entity settings.

The default is also applied where the value is read, not only at startup. `errgroup.SetLimit(0)` does not mean unbounded — it admits no goroutines and `Wait` blocks forever — so an unset value has to mean "use the default" wherever it comes from. `Serve` fills it in, but an API built without being served keeps the zero, which is every test in the package.

**Panic recovery.** Correct, and it was a regression: the serial version was covered by the recover in `GoRoutines.exec` further up. The read now recovers, logs the stack the same way, and returns the panic as the error the group already propagates.

**Cancellation after a failure.** Also correct. The loop breaks on `groupCtx.Err() != nil`.

**Comments.** Cut to what the code does.

Tests, against a live Postgres with the VCS service mocked:

- bound set to 1 must serialise the reads — this is what proves the configured value is read rather than the default;
- unset bound must complete rather than deadlock, asserted as a call that finishes, since the defect is a hang;
- a panicking read must fail the analysis, not the process (without the recover this does not fail, it takes the test binary down);
- a failure part way through a 200-file tree must not read all 200.

Separately: on your comment in #7834 that 1s at p50 is too much, you were right and the cause was on our side, not the forge. Written up there. It does not change this PR — the fan-out is still one round trip per entity in series — but it does change how much of the 5m30s it was responsible for.

engine/api/v2_repository_analyze.go
gsalingu: Done in f2f0434: `entity.fileFetchConcurrency`, default 8, applied in `Serve` next to the other entity settings.

The default is also applied where the value is read, not only at startup — `errgroup.SetLimit(0)` admits no goroutines and `Wait` then blocks forever, so an unset bound has to mean the default rather than a hang. A test covers each half.

engine/api/v2_repository_analyze.go
gsalingu: Removed in f2f0434.

engine/api/v2_repository_analyze.go
gsalingu: Removed in f2f0434. 8d04963 then went over every remaining comment in this PR the same way — what the code does, with the measurements left in the PR description.

engine/api/v2_repository_analyze.go
gsalingu: Both fixed in 0dcacae.

1. The read now recovers, logs the stack the way `GoRoutines.exec` does, and returns the panic as the error the group already propagates. Covered by a test that does not fail without the recover — it ends the test binary, which is the same thing happening in miniature.

2. The loop breaks on `groupCtx.Err() != nil`. Covered by a test asserting a failure part way through a 200 file tree does not read all 200.

sguiheux (approved):

PR #7839, opened 2026-09-03
perf(hooks): run the next state of a hook event without waiting for the sweep

### LINKED ISSUE #7835: Hook events advance one state per sweep, adding minutes between a push and its run
## What happens

`executeEvent` in `engine/hooks/scheduler_repository_event.go` advances a hook
event by at most one state per dequeue, and nothing re-queues it afterwards:

```go
switch hre.Status {
case sdk.HookEventStatusScheduled:  ... // -> Analysis
case sdk.HookEventStatusAnalysis:   ... // -> WorkflowHooks
case sdk.HookEventStatusWorkflowHooks: ... // -> GitInfo
case sdk.HookEventStatusGitInfo:    ... // -> Workflow
case sdk.HookEventStatusWorkflow:   ...
}
return nil   // event is left in the in-progress list
```

The event then waits to be picked up by `manageOldRepositoryEvent`, which ticks
once every `OldRepositoryEventRetry` minutes (default 1):

```go
tick := time.NewTicker(time.Duration(s.Cfg.OldRepositoryEventRetry) * time.Minute).C
```

A push walks `Scheduled → Analysis → WorkflowHooks → GitInfo → Workflow`, so it
pays that interval four or five times over, while the work in each state takes
seconds.

## Measurements

The queue is not the problem — it was empty throughout:

```
Queue                     0
BalanceRepositoryEvent    769 in / 863 out
dequeueRepositoryEvent    Running
```

Measured on one push before the change: the analysis completed **2 seconds**
after the event was created, and the workflow run appeared **2 minutes 49
seconds** after that. The time was spent waiting for sweeps, not working.

After re-queueing on state change, traced per transition:

```
+6s   Analyzing
+31s  GitInfo
+36s  Done
```

## Proposed fix

Re-queue the event as soon as a pass advances it, so the states chain instead of
waiting for the sweep.

Only when the status actually changed. A trigger that returns without advancing
is waiting on something it does not own — an analysis still running, an
operation outstanding — and re-queueing that spins the worker instead of waiting
for the callback that resolves it. We tried the unconditional version and it
measured *worse* (94 s → 141–182 s) because the dequeue loop is serial and
parked events crowd out real work. The periodic sweep stays as the backstop.

PR: https://github.com/ovh/cds/pull/7839

gsalingu:
1. Description

`executeEvent` advances a hook event by at most one state per dequeue and
nothing re-queued it afterwards, so each transition waited for
`manageOldRepositoryEvent` — a ticker of `OldRepositoryEventRetry` minutes,
default 1. A push walks Scheduled -> Analysis -> WorkflowHooks -> GitInfo ->
Workflow, so it paid that interval four or five times over while the work in
each state took seconds.

This was not congestion. The queue was empty throughout (`Queue 0`, dequeue
goroutines running, more events out than in). On one measured push the analysis
completed 2 seconds after the event was created, and the run appeared 2 minutes
49 seconds later.

The event is now re-queued as soon as a pass advances it, so the states chain.

Only when the status actually changed. A trigger that returns without advancing
is waiting on something it does not own — an analysis still running, an
operation outstanding — and re-queueing that spins the worker rather than
waiting for the callback that resolves it. We tried the unconditional form and
measured it worse (94s -> 141s and 182s): the dequeue loop is serial, so parked
events crowd out real work. The periodic sweep remains the backstop for states
that are genuinely waiting.

Traced per transition after the change: +6s Analyzing, +31s GitInfo, +36s Done.

1. Related issues

Closes #7835

1. About tests

Four tests in `engine/hooks/scheduler_repository_event_test.go`, against a live
Redis. Both sides of the condition are covered, and each side fails if removed:

- `TestExecuteEventRequeuesItselfWhenItAdvances` — a push event at Scheduled
  advances to Analysis and must be back on the queue. Red without the re-queue.
- `TestExecuteEventDoesNotRequeueWhileWaiting` — an event at Analysis whose
  analysis is still running does not advance and must not be re-queued. Red
  against the unconditional form described above, which is the variant we tried
  first and measured worse.
- `TestExecuteEventDoesNotRequeueTerminalEvents` — a finished event must leave
  the queue rather than be put back on it.
- `TestIsTerminalHookEventStatus` — pins which states end an event. Getting this
  wrong is silent in either direction: a terminal state treated as intermediate
  is re-queued forever, an intermediate state treated as terminal stalls exactly
  like the bug this fixes.

Queue assertions are relative rather than absolute, since the queue is shared.

The full `engine/hooks` package suite passes (29 tests) against a live Redis,
which is also where a re-queue loop or goroutine leak would surface.

Verified in production for several hours: no growth in the event queue, no
repeated processing of the same event, and push-to-run latency reduced as above.

1. Mentions

@ovh/cds

### DIFF
--- a/engine/hooks/scheduler_repository_event.go
+++ b/engine/hooks/scheduler_repository_event.go
@@ -163,6 +163,10 @@ func (s *Service) manageRepositoryEvent(ctx context.Context, eventKey string) er
 }
 
 func (s *Service) executeEvent(ctx context.Context, hre *sdk.HookRepositoryEvent) error {
+	// One pass advances the event by at most one state. Whether the next state
+	// runs now or waits for the sweep is decided at the end of this function.
+	statusOnEntry := hre.Status
+
 	ctx, next := telemetry.Span(ctx, "s.executeEvent")
 	defer next()
 	defer func() {
@@ -230,5 +234,33 @@ func (s *Service) executeEvent(ctx context.Context, hre *sdk.HookRepositoryEvent
 			log.Error(ctx, "executeEvent >unable to remove event %s from inprogress list: %v", hre.UUID, err)
 		}
 	}
+
+	// Run the next state now rather than waiting to be swept up.
+	//
+	// A push walks Scheduled -> Analysis -> WorkflowHooks -> GitInfo ->
+	// Workflow, and each pass through this function handles exactly one of
+	// them. Nothing re-queued the event in between, so every transition waited
+	// for manageOldRepositoryEvent, which ticks once per
+	// OldRepositoryEventRetry minute: four or five states meant minutes between
+	// a push and its run while the work itself took seconds.
+	//
+	// Only re-queue when this pass actually moved the event. A trigger that
+	// returns without advancing is waiting on something it does not own — an
+	// analysis still running, an operation outstanding — and re-queueing that
+	// would spin the worker instead of waiting for the callback that resolves
+	// it. The periodic sweep stays as the backstop for exactly those.
+	if hre.Status != statusOnEntry && !isTerminalHookEventStatus(hre.Status) {
+		if err := s.Dao.EnqueueRepositoryEvent(ctx, hre); err != nil {
+			return sdk.WrapError(err, "unable to enqueue repository event %s for its next state", hre.GetFullName())
+		}
+	}
 	return nil
+}
+
+func isTerminalHookEventStatus(status string) bool {
+	switch status {
+	case sdk.HookEventStatusDone, sdk.HookEventStatusSkipped, sdk.HookEventStatusError:
+		return true
+	}
+	return false
 }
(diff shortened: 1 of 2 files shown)


engine/hooks/scheduler_repository_event.go
```diff
@@ -230,5 +234,33 @@ func (s *Service) executeEvent(ctx context.Context, hre *sdk.HookRepositoryEvent
 			log.Error(ctx, "executeEvent >unable to remove event %s from inprogress list: %v", hre.UUID, err)
 		}
 	}
+
+	// Run the next state now rather than waiting to be swept up.
```
sguiheux, line 238: Hi, it's wrong, each state trigger the next one, so we don"t need to re-enqueue the event
manageOldRepositoryEvent trigger is here to  catch lost event and it's not used in a nominal case

Maybe you have issue on receiving callbacks ( analysis step/ git info step )

gsalingu: Thanks for the comment I'll take a look on my side!

Issue #7836, opened 2026-09-03
Repository analyses stuck InProgress silently starve the poller


gsalingu:
## What happens

A repository analysis can stop being processed entirely, with no error, no
timeout and nothing in the logs. Two things combine to cause it.

**The poller's window is unordered.**
`engine/api/repository/dao_vcs_project_repository_analyze.go`:

```go
query := gorpmapping.NewQuery("SELECT * FROM project_repository_analysis WHERE status = $1 LIMIT 100")
```

**Nothing ever terminates an analysis that cannot finish.**
`cleanRepositoryAnalysis` trims per-repository retention only, and there is no
deadline anywhere on an in-progress analysis.

So an analysis whose owner died — API restarted mid-flight, hook event gone,
lock holder vanished — stays `InProgress` for the life of the database. Enough
of them and they occupy the hundred rows the poller reads, and every analysis
created afterwards is invisible to it. Not being looked at is indistinguishable
from not being finished, which is why this surfaces as silence rather than as
failure.

## Measurements

On our instance, `project_repository_analysis` reached:

```
status = InProgress                    503
   of which older than 30 minutes      483
oldest                                 dating from the start of the incident
```

The tell was that manually triggered analyses on a repository stopped completing
entirely while webhook-driven analyses on the *same repository, minutes apart*
still succeeded. Nothing distinguished them except which side of the hundred-row
window they landed on.

Downstream this presented as pushes producing no CI run at all: a push created
an analysis, the analysis was never picked up, and the hook event behind it sat
in `Analyzing` indefinitely.

## Our approach, and where we would like your view

We treated these as two separable problems and implemented both. They are
proposed as two PRs so the low-risk half can be judged on its own.

### 1. Make the poller resilient to a backlog — `ORDER BY created DESC`

PR: https://github.com/ovh/cds/pull/7840

One line. Fresh work stays reachable however large the backlog grows, and there
is no behaviour change on a healthy instance. We think this is
uncontroversial and would stand alone even if you reject the second part.

### 2. Stop the backlog forming — a deadline on in-progress analyses

PR: https://github.com/ovh/cds/pull/7841

A reaper on the existing poller tick: anything `InProgress` past 30 minutes is
failed with an explanatory message, 200 per sweep so a large backlog drains over
several ticks rather than in one long transaction.

On deploy it logged `expiring 196 repository analyses left in progress for more
than 30m0s`; the backlog went 472 -> 46 and older-than-30-minutes to zero.

[...]

PR #7840, opened 2026-09-03
fix(api): keep the newest analyses visible to the poller

gsalingu:
1. Description

The analysis poller selects its work with `LIMIT 100` and no `ORDER BY`. An
analysis whose owner died stays `InProgress` indefinitely, so those rows can
occupy the hundred the poller reads and analyses created afterwards are never
returned — with no error, no timeout and nothing in the logs.

On our instance the table reached 503 `InProgress` rows, 483 of them beyond half
an hour old. While that was true, manually triggered analyses on a repository
stopped completing entirely while webhook-driven analyses on the same repository
minutes apart still succeeded — the only difference being which side of the
window they fell on.

`ORDER BY created DESC` keeps fresh work reachable however large the backlog
grows. No behaviour change on a healthy instance.

1. Related issues

Part of #7836

Related: the second PR on that issue — this keeps the system usable when a
backlog exists but does not stop one forming. That change is more opinionated and is proposed
separately so this one can be judged on its own.

1. About tests

`TestAnalysisInProgressPollerSeesNewestWorkPastABacklog` in
`engine/api/v2_repository_analyze_test.go`, against a live Postgres. It builds a
backlog of 120 InProgress analyses, explicitly aged so the ordering is the only
thing deciding what comes back, then creates one more and asserts the poller's
window still contains it.

It is a real regression test rather than a restatement of the code: reverting
the `ORDER BY` alone turns it red — the newest analysis is absent from the
hundred rows returned — and green again with it.

too).

1. Mentions

@ovh/cds

### DIFF
--- a/engine/api/repository/dao_vcs_project_repository_analyze.go
+++ b/engine/api/repository/dao_vcs_project_repository_analyze.go
@@ -112,8 +112,16 @@ func LoadAnalysesByRepo(ctx context.Context, db gorp.SqlExecutor, projectReposit
 	return getAnalyses(ctx, db, query)
 }
 
+// LoadRepositoryIDsAnalysisInProgress returns the analyses the poller should
+// pick up next.
+//
+// The ordering matters. This is a fixed-size window over every row sitting in
+// InProgress, and an analysis whose owner died stays in that state, so without
+// an order the window fills with whichever hundred rows the table happens to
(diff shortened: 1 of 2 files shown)


sguiheux:
As explained in the ticket I don't think it's the right fix

PR #7841, opened 2026-09-03
fix(api): expire repository analyses left in progress

gsalingu:
1. Description

**Proposed for discussion rather than merge as-is — please see the questions in
the issue.**

Nothing terminates a repository analysis that cannot finish.
`cleanRepositoryAnalysis` trims per-repository retention only, and there is no
deadline on an in-progress analysis, so one whose owner died stays `InProgress`
for the life of the database and permanently consumes a poller slot.

This adds a reaper on the existing poller tick: anything `InProgress` past 30
minutes is failed with an explanatory message, 200 per sweep so a large backlog
drains over several ticks rather than in one long transaction.

On deploy it logged `expiring 196 repository analyses left in progress for more
than 30m0s`; the backlog went 472 -> 46 and older-than-30-minutes to zero.

We are not confident this is the shape you want. Specifically: whether 30
minutes is right and should be configurable; whether expiry belongs here or in
`cleanRepositoryAnalysis`; whether `Error` is the right terminal state or an
expired analysis deserves its own; and whether you would rather fix whatever
orphans analyses in the first place than reap them. Ours were created during an
API crash-loop and we did not chase that root cause.

Happy to rework or withdraw this.

1. Related issues

Part of #7836

Complements the ordering PR on the same issue, which makes the poller resilient
to a backlog while this one stops the backlog forming.

1. About tests

Two tests in `engine/api/v2_repository_analyze_test.go`, against a live Postgres:

- `TestExpireStaleRepositoryAnalysesFailsOnlyThePastDeadlineOnes` — both
  directions of the deadline, which is where the risk sits. An analysis past the
  deadline ends in `Error` carrying a message that says what happened; one
  inside the deadline is left alone.
- `TestLoadRepositoryAnalysesStuckInProgressRespectsItsBound` — the sweep
  honours its limit and returns oldest first, so a large backlog drains from its
  head over several ticks rather than in one long transaction.

Timestamps are backdated in SQL, which is safe here: the signed canonical form
is `{ID}{ProjectRepositoryID}{VCSProjectID}{ProjectKey}{Commit}`, so timestamps
and status sit outside the signature.

Also verified in production against a real backlog of 503 rows.

too).

If the shape changes after the discussion in the issue, these move with it.

1. Mentions

@ovh/cds

### DIFF
--- a/engine/api/repository/dao_vcs_project_repository_analyze.go
+++ b/engine/api/repository/dao_vcs_project_repository_analyze.go
@@ -117,6 +117,15 @@ func LoadRepositoryIDsAnalysisInProgress(ctx context.Context, db gorp.SqlExecuto
 	return getAnalyses(ctx, db, query)
 }
 
+// LoadAnalysesInProgressIdleSince returns up to limit InProgress analyses not
+// modified since the given time, oldest first.
+func LoadAnalysesInProgressIdleSince(ctx context.Context, db gorp.SqlExecutor, idleSince time.Time, limit int64) ([]sdk.ProjectRepositoryAnalysis, error) {
+	query := gorpmapping.NewQuery(
+		"SELECT * FROM project_repository_analysis WHERE status = $1 AND last_modified < $2 ORDER BY last_modified ASC LIMIT $3",
+	).Args(sdk.RepositoryAnalysisStatusInProgress, idleSince, limit)
+	return getAnalyses(ctx, db, query)
+}
+
 func LoadRepositoryAnalysisById(ctx context.Context, db gorp.SqlExecutor, projectRepoID, analysisID string) (*sdk.ProjectRepositoryAnalysis, error) {
 	ctx, next := telemetry.Span(ctx, "repository.LoadRepositoryAnalysisById")
 	defer next()
--- a/engine/api/v2_repository_analyze.go
+++ b/engine/api/v2_repository_analyze.go
@@ -357,7 +357,7 @@ func (api *API) createAnalyze(ctx context.Context, tx gorpmapper.SqlExecutorWith
 	return &repoAnalysis, nil
 }
 
-func (api *API) repositoryAnalysisPoller(ctx context.Context, tick time.Duration) {
+func (api *API) repositoryAnalysisPoller(ctx context.Context, tick time.Duration, analysisTimeout time.Duration) {
 	ticker := time.NewTicker(tick)
 	defer ticker.Stop()
 
@@ -369,6 +369,7 @@ func (api *API) repositoryAnalysisPoller(ctx context.Context, tick time.Duration
 			}
 			return
 		case <-ticker.C:
+			api.expireStaleRepositoryAnalyses(ctx, analysisTimeout)
 			analysis, err := repository.LoadRepositoryIDsAnalysisInProgress(ctx, api.mustDB())
 			if err != nil {
 				log.Error(ctx, "unable to load analysis in progress: %v", err)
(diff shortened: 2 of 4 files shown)


engine/api/v2_repository_analyze.go
```diff
@@ -390,6 +391,56 @@ func (api *API) repositoryAnalysisPoller(ctx context.Context, tick time.Duration
 	}
 }
 
+// repositoryAnalysisDeadline bounds how long an analysis may sit InProgress.
+// An analysis is seconds of work; anything still running after this lost the
+// worker that owned it and no path brings it back.
+const repositoryAnalysisDeadline = 30 * time.Minute
```
sguiheux: Could you move that in the CDS configuration, and manage a default value in the code ( like the others CDS parameters

engine/api/v2_repository_analyze.go
```diff
@@ -390,6 +391,56 @@ func (api *API) repositoryAnalysisPoller(ctx context.Context, tick time.Duration
...
+
+// repositoryAnalysisExpiryBatch bounds one sweep so a large backlog drains over
+// several ticks rather than in one long transaction.
+const repositoryAnalysisExpiryBatch = int64(200)
```
sguiheux: Same as above

engine/api/repository/dao_vcs_project_repository_analyze.go
```diff
@@ -117,6 +117,16 @@ func LoadRepositoryIDsAnalysisInProgress(ctx context.Context, db gorp.SqlExecuto
 	return getAnalyses(ctx, db, query)
 }
 
+// LoadRepositoryAnalysesStuckInProgress returns analyses left InProgress past
+// their deadline, oldest first, so they can be failed rather than occupying
+// poller capacity for the life of the process.
+func LoadRepositoryAnalysesStuckInProgress(ctx context.Context, db gorp.SqlExecutor, before time.Time, limit int64) ([]sdk.ProjectRepositoryAnalysis, error) {
```
sguiheux: Could you renamed the function as LoadAnalysesInProgressIdleSince ( or something similar )
with a comment like: 
// LoadAnalysesInProgressIdleSince returns up to limit InProgress analyses not
// modified since the given time, oldest first.

We don't want your bug context in the comment

engine/api/v2_repository_analyze.go
gsalingu: Will do!

engine/api/repository/dao_vcs_project_repository_analyze.go
gsalingu: Will do!

gsalingu:
Thanks — all three applied.

**Configuration.** `analysisTimeout` (default `30m`) and `analysisExpiryBatch` (default `200`) are now under `[entity]`, parsed in `Serve` next to the existing `retention`.

One deliberate difference from the other parameters: the defaults are also applied where the values are read, not only at startup. A zero timeout would expire every in-progress analysis on the next tick, and I recently spent a while on a config value that reached `errgroup.SetLimit(0)` — which admits no goroutines and blocks forever — because it never went through `Serve`. Happy to drop the second check if you'd rather it match the others exactly.

The Go field names are shorter than the toml keys (`AnalysisTimeout`, `AnalysisBatch`) so the struct alignment of the existing `[entity]` fields is untouched. On that note, `engine/api/api.go` isn't gofmt-clean on master — three `Workflow` fields — which is why I kept my formatting to my own lines rather than running gofmt over the file.

**Naming.** Renamed to `LoadAnalysesInProgressIdleSince` with your comment, and I've taken the incident narrative out of the other doc comments too.

Squashed to a single commit. Apologies — I saw the "don't amend during review" note in CONTRIBUTING only afterwards, so your inline comments now point at replaced commits; the changes above are the whole of what moved.

Full `engine/api` package suite green against Postgres and Redis.

engine/api/v2_repository_analyze.go
```diff
@@ -390,6 +391,62 @@ func (api *API) repositoryAnalysisPoller(ctx context.Context, tick time.Duration
...
+const (
+	defaultAnalysisTimeout = 30 * time.Minute
+	defaultAnalysisBatch   = int64(200)
+)
+
+// expireStaleRepositoryAnalyses fails analyses that outlived the timeout.
+//
+// Nothing else does: cleanRepositoryAnalysis only trims per-repository
+// retention, so an analysis that cannot finish stays InProgress for the life
+// of the database and permanently consumes one of the slots the poller reads.
```
sguiheux, line 405:
```suggestion
```

sguiheux (changes requested):
Just a little comment.
Please do not force push, it allows us to review commit by commit

gsalingu:
Suggestion applied in a new commit. No force push — apologies for the last one.

One thing you should know before spending more time on this. Your question in #7836 was whether a reaper is a plaster over a different bug. Having chased it since, I think it partly is, and the answer changes what I would propose here.

Two things orphan analyses on our instance:

1. The per-analysis lock (`api:analyzeRepository:<id>`, 5 minute TTL) is taken for the length of the work and never refreshed, so any analysis running longer than the TTL loses its lock while it is still running. Everything downstream then reads it as unowned.
2. Nothing drains in-flight analyses at shutdown, so an API restart leaves every running analysis `InProgress` with no worker behind it. Ours were created during a crash-loop, which is why there were 500 of them.

With those two fixed, the remaining orphans are recoverable rather than fatal, so we no longer fail them. The sweep asks whether the per-analysis lock is still held: if it is, the analysis is slow, not stuck, and is left alone; if it is not, the row is handed back to the queue, bounded at three attempts before it is failed for real. Failing an analysis that was never attempted is the wrong outcome — it is not a failed analysis, it is an unstarted one, and it is a push that silently loses its build.

That shape depends on the `ToStart` status you suggested in #7836, which is why I have not rewritten this PR into it.

So: happy to leave this as the plain deadline you are reviewing, reshape it into the lock-check plus requeue, or close it if you would rather fix the orphaning directly and not carry a reaper at all. Whichever you prefer.

engine/api/v2_repository_analyze.go
gsalingu: Applied in 38ad0d9 — the doc comment is the one line saying what the function does. 7918b02 then went over the rest of this PR the same way.

engine/api/v2_repository_analyze.go
gsalingu: Done: `entity.analysisTimeout` (30m) and `entity.analysisExpiryBatch` (200), parsed in `Serve` next to the existing retention.

engine/api/v2_repository_analyze.go
gsalingu: Same.

engine/api/repository/dao_vcs_project_repository_analyze.go
gsalingu: Renamed to `LoadAnalysesInProgressIdleSince`, with your comment as written.

sguiheux (approved):

sguiheux:
Ok for me, CI: OK
Will wait for a second review before merging

yesnault (approved):

Issue #7847, opened 2026-09-07
RFC: VCS provider clients are built per request, so the whole forge fan-out shares one connection


gsalingu:
## Summary

A VCS provider client is built per request — every handler in `engine/vcs` calls `GetAuthorizedClient` — and nothing in that construction is shared. Two costs follow, both measurable, both invisible in logs:

1. **All forge traffic goes down one TCP connection.** The clients carry a nil `Transport`, so they share `http.DefaultTransport`. Against an HTTP/2 forge, Go's h2 client keeps one connection per host and multiplexes onto it, so a repository analysis reading its entity files concurrently serialises behind that one connection.
2. **Every Gitea request costs an extra round trip.** `gg.NewClient` probes `/api/v1/version` behind a `sync.Once` held on the *client instance*. A client per request means a `sync.Once` per request, so the probe fires every time.

This is an RFC rather than a plain bug report: the first has several plausible fixes with different trade-offs, and the second cannot be fixed without giving something up. PR attached, but please read it as a proposal.

### What we measured

On our instance, with a Forgejo forge and repositories carrying up to ~130 `.cds` entity files:

```
netstat inside a live cds-vcs pod, during a 3-repository analysis burst
  ESTABLISHED connections to the forge     1
  peak in-flight requests (10 min window)  760

/content latency, same pod, same forge
  n=637  p50=1237ms  p90=2295ms  max=4422ms
  same endpoint, same basic auth, same commit ref, one curl from that pod:  60ms

cds-vcs itself: 5m CPU, no throttling
```

The forge, the network, DNS, the auth scheme and the ref form were each measured and eliminated. What is left is that concurrency is not reaching the wire.

### Reproduction

Stdlib only, Go 1.26, an HTTP/2 `httptest` server counting distinct connections under a 20-way burst, lanes warmed first:

```
shared transport (what CDS does today)     20 requests -> 1 TCP connection
shared transport + MaxConnsPerHost=8       20 requests -> 1 TCP connection
pool of 8 transports                       20 requests -> 8 TCP connections
```

The same assertion is in the PR as a test, against a real h2 `httptest` server.

### What does not fix it

Each of these looks like the answer and is not:

- **`MaxIdleConnsPerHost`** (2 on `DefaultTransport`) — a genuine constraint for an HTTP/1.1-only forge, irrelevant once h2 is negotiated.
[...]

PR #7848, opened 2026-09-07
fix(vcs): give the forge several connections instead of one

gsalingu:
1. Description

**Proposed for discussion rather than merge as-is — the open questions are in #7847.**

A VCS provider client is built per request: every handler in `engine/vcs` calls `GetAuthorizedClient`. Those clients carry a nil `Transport`, so all of them share `http.DefaultTransport`, and against an HTTP/2 forge Go's h2 client keeps one connection per host and multiplexes onto it. A repository analysis reading its entity files concurrently therefore serialises behind a single connection however wide the fan-out is.

Measured on our instance, Forgejo forge, repositories carrying up to ~130 `.cds` entity files:

```
netstat inside a live cds-vcs pod during a 3-repository analysis burst
  ESTABLISHED connections to the forge     1
  peak in-flight requests (10 min window)  760

/content latency, same pod
  n=637  p50=1237ms  p90=2295ms  max=4422ms
  one curl, same endpoint, same auth, same ref:  60ms

cds-vcs: 5m CPU, no throttling
```

This adds `engine/vcs/transport`, holding a small pool of transports that outlive a request, handed out in rotation — the per-request part of a client is its credentials, not its connection pool.

Four things that look like the fix and are not, since they will come up in review: `MaxIdleConnsPerHost` (2 on `DefaultTransport`) matters for an HTTP/1.1-only forge and not once h2 is negotiated; giving each client its own `Transport` is strictly worse, because the client is per request so nothing is pooled at all; `MaxConnsPerHost` bounds how many connections a cold dial race may produce, not whether a warm healthy connection gets siblings; forcing HTTP/1.1 does work, by giving up multiplexing and header compression.

The second commit wires the Gitea provider the same way and gives it a timeout — the SDK builds `&http.Client{}` with neither. It deliberately does **not** remove that provider's `/api/v1/version` probe, which fires on every request because the SDK's `sync.Once` lives on a client CDS rebuilds each time. Both ways to remove it trade something away; the choice is in #7847.

`github`, `gitlab`, `bitbucketcloud` and `bitbucketserver` have the same per-request construction and are untouched: we have measured only Forgejo, and did not want to change six providers on one instance's evidence.

1. Related issues

Part of #7847
Explains the p50 discussed in #7834

1. About tests

Two in `engine/vcs/transport/transport_test.go`, both mutation-checked against the change they cover:

- `TestPooledRotates` — the pool must hand out every transport it holds. Pinned directly, because without the rotation the pool is decoration: every request would go through one transport, which is the state this package exists to leave. Returning `pool[0]` turns it red.
- `TestPooledOpensSeveralConnectionsToAnHTTP2Host` — the reason the package exists, against a real HTTP/2 `httptest` server counting distinct connections under a warmed twenty-four way burst. One transport answers it on one connection; the pool answers it on several.

The lanes are warmed before the burst on purpose: a transport dialling a cold host does open one connection per concurrent request, but that is a dial race rather than steady state, and a long-running service is never in it. Without warming the test would pass for the wrong reason.

The first assertion is deliberately written as a claim about Go rather than about this package. If it ever stops holding, this package has no reason to exist, and the test should be the thing that says so.

`go build ./engine/...` and `go test ./engine/vcs/transport/` pass. `go vet ./engine/vcs/...` reports one pre-existing unkeyed-fields warning in `gitea/client_branch.go`, unchanged by this PR.

1. Mentions

@ovh/cds

### DIFF
--- a/engine/vcs/forgejo/http_client.go
+++ b/engine/vcs/forgejo/http_client.go
@@ -10,6 +10,8 @@ import (
 	"net/url"
 	"strings"
 	"time"
+
+	"github.com/ovh/cds/engine/vcs/transport"
 )
 
 // forgejoHTTPClient is a simple HTTP client for the Forgejo API.
@@ -26,7 +28,7 @@ func newForgejoHTTPClient(baseURL, username, token string) *forgejoHTTPClient {
 		baseURL:    strings.TrimRight(baseURL, "/") + "/api/v1",
 		username:   username,
 		token:      token,
-		httpClient: &http.Client{Timeout: 60 * time.Second},
+		httpClient: &http.Client{Timeout: 60 * time.Second, Transport: transport.Pooled()},
 	}
 }
 
--- a/engine/vcs/gitea/gitea.go
+++ b/engine/vcs/gitea/gitea.go
@@ -3,11 +3,14 @@ package gitea
 import (
 	"context"
 	"fmt"
+	"net/http"
 	"strings"
+	"time"
 
 	gg "code.gitea.io/sdk/gitea"
 
 	"github.com/ovh/cds/engine/cache"
+	"github.com/ovh/cds/engine/vcs/transport"
 	"github.com/ovh/cds/sdk"
 )
 
@@ -56,7 +59,13 @@ func New(URL, apiURL, uiURL, proxyURL string, store cache.Store, username, token
 
 // GetAuthorizedClient returns an authorized client
 func (g *giteaConsumer) GetAuthorizedClient(_ context.Context, vcsAuth sdk.VCSAuth) (sdk.VCSAuthorizedClient, error) {
-	client, err := gg.NewClient(g.URL, gg.SetBasicAuth(g.username, g.token))
+	client, err := gg.NewClient(g.URL,
+		gg.SetBasicAuth(g.username, g.token),
+		// This client is built per request, so its transport comes from the
+		// process-wide pool. The SDK's own default has neither a transport
+		// nor a timeout.
+		gg.SetHTTPClient(&http.Client{Timeout: 60 * time.Second, Transport: transport.Pooled()}),
+	)
 	if err != nil {
 		return nil, sdk.WithStack(err)
 	}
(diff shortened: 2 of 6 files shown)


gsalingu:
Made the pool configurable up front rather than asking, since the same question came up on #7838 and the answer there applies unchanged: how much concurrency a forge will take is an operator's knowledge, not something CDS can see.

`vcs.forge.connectionPoolSize` (default 8) and `vcs.forge.maxIdleConnsPerHost` (default 32), applied in `ApplyConfiguration` before any handler can build a provider client.

Same treatment as `entity.fileFetchConcurrency` for the unset case: the defaults are applied where the values are read, not only where they are configured. A pool sized straight from an absent value holds no transports at all, so the first forge request of a service that had simply not set the knob would panic on an empty slice — an unset value means the default, not none. That half of the test is mutation-checked: dropping the clamp turns it red.

The pool is swapped wholesale rather than mutated, so a request already choosing a lane cannot observe a half-built one.

One thing I did decide and would like your view on: the knobs sit at the VCS **service** level (`[vcs.forge]`) rather than per provider, because the pool is shared by Forgejo and Gitea today and would be shared by the other four. If you would rather each provider carried its own, that is a small change.

Question 3 in #7847 is struck through accordingly; the other three still stand.

sguiheux:
Streams on an h2 connection are multiplexed, and the server announces how many it accepts per connection: 250 for a Go server such as Forgejo. Below that limit, the forge receives every request at the same time on that one connection, and handles each stream in its own goroutine. Above it, the default Go client opens a second connection on its own. So "serialises behind a single connection" only holds at the TCP level (bandwidth, retransmission), which for small YAML files should not be visible, and the 5m CPU on cds-vcs says the client read loop is not the bottleneck either.

Checked on Go 1.26 with a plain cloned `http.DefaultTransport`, no pool: against an h2 server capped at 2 streams per connection, a burst of 10 requests opened 6 TCP connections on its own. The default client already dials more connections when a connection is full. 

Your netstat showing exactly one established connection therefore means the forge's stream limit was never reached, so every in-flight request was already reaching the forge concurrently.

PR #7859, opened 2026-09-18
perf(api): stop re-walking signed run results on every synchronize

yesnault:
The worker calls PUT /v2/queue/{region}/job/{id}/runresult/synchronize with a 30s HTTP timeout, and synchronizeRunResults does all its Artifactory work inside that request. Every call walks all the results of the run again, including the ones signed by an earlier call, at three Artifactory round trips each (GetRepository, GetFileInfo, GetProperties). Past a certain number of artifacts the handler goes over 30s and the worker fails the step while the API is still working, which is what happens on actions/artifactoryRelease.

An already signed result now costs a single call: properties are read before file info, and the loop stops there when cds.signature is present. Repository details are read once per repository rather than once per result, and loading the run job moved after the signature check.


fsamin (approved):

yesnault:
https://github.com/ovh/cds/pull/7870

